PR #21, opened 2026-04-26
examples: add claude-code-headless + agent-web-ui CC Exec view

M64GitHub:
## Summary

Adds two co-evolved pieces:

- **`examples/claude-code-headless`** — a NATS-driven service that spawns Claude Code sessions on demand via `@anthropic-ai/claude-agent-sdk`. Each session registers as a first-class NATS Agent Protocol v0.2.0-draft instance at `agents.cc.<owner>.<session_id>`; a small controller at `agents.cc.<owner>.<name>` exposes `spawn` / `stop` / `list` endpoints. Streams text per token, surfaces tool calls + results inline, asks for permission via §7 query chunks, tracks per-turn and cumulative cost.
- **`examples/agent-web-ui`** — new "CC Exec" mode mirroring the existing PI Exec workspace. Controller selector + spawn form (cwd, model, allowed tools, permission mode, max turns, max lifetime) on the left; chat surface on the right with collapsible tool cards (name + input + result + success/error glyph) and Allow/Deny prompts when Claude wants a tool that isn't pre-approved. Per-turn cost on each agent bubble; cumulative cost + turn count on each session card.

Together they're a working "headless Claude Code in a browser tab" demo: spawn a session over NATS, prompt it, watch Claude type, approve tool calls one by one, see the running cost — all driven by clicks in a Vue 3 SPA backed by a Bun WebSocket bridge.

## What's in the diff

**Server side** (`examples/claude-code-headless/`):
- Controller registers a `cc`-token NATS micro service, exposes `spawn` / `stop` / `list` endpoints + a `prompt` endpoint that returns help text, publishes 30s heartbeats.
- Per-session bridge owns one Claude Agent SDK lifecycle behind a `ReferenceAgent`. Serial drain queue, ` query()` with `includePartialMessages`, cancellation via `AbortController`, session resume via `resume: <sdkSessionId>`.
- Permission flow: `canUseTool` callback emits a fresh `query` chunk on a unique inbox + awaits one reply (2 min timeout); `yes/y/allow/approve/ok` → allow, anything else → deny with reply preserved as the reason. Echoes the tool input back as `updatedInput` because the SDK's runtime Zod validator requires it (TS type marks it optional — mismatch noted below).
- Tool observability + cost as prefix-tagged status payloads (`tool_use:<json>`, `tool_result:<json>`, `cost:<json>`). Spec-compliant — `status.data` stays a string per §6.4. SDK callers without prefix knowledge see opaque status tokens.
- Resolves the bundled `claude` binary via CLI flag → env → config file → `Bun.which("claude")` → SDK fallback. Side-steps the SDK's per-platform-package auto-detect picking a variant that isn't installed.
- Conservative defaults: `permissionMode: "dontAsk"`, `allowedTools: [Read, Glob, Grep]`, `claude-sonnet-4-6`, 50-turn cap, 30-min lifetime — overridable per-spawn.

**Client side** (`examples/agent-web-ui/`):
- Bridge handlers translate the prefix-tagged status payloads into typed `tool-use` / `tool-result` / `cost` server messages.
- Generic `controlRequest<T>` helper replaces the PI-specific `piexecRequest`; CC Exec endpoints share the routing.
- New `tool` MessageRole + ToolCallInfo data model; `findMessageByToolId` pairs results back with their tool_use bubble.
- New CC Exec view (`src/components/ccexec/`): ControllerSelector, SpawnForm, SessionList, CcExecView. Mirrors PI Exec's shape; reuses the generic chat / message-list / prompt-area surface.
- App.vue: third "CC Exec" mode toggle, auto-shown when a controller is discovered, vanish-fallback to chat mode when it disappears.
- Drive-by fix: `crypto.randomUUID()` polyfill for non-secure HTTP origins (`src/uuid.ts`). Both PI Exec and CC Exec hit this when accessed from a LAN browser; without the polyfill the click handler died silently.

**Shared docs** (READMEs for both `examples/claude-code-headless` and `agents/claude-code`):
- API-key-only auth posture: `ANTHROPIC_API_KEY` recommended for both Claude Code wrappers regardless of deployment scope. The runtime fallback to `~/.claude` still works (the SDK uses the env var in preference when both are present), but the recommended path in docs is unambiguous to avoid the personal-use-vs-multi-user trap.

## Out of scope (roadmap)

Documented in the headless README's Roadmap section:
- Programmatic MCP server attachment per spawn (`mcp_servers` field).
- System prompt / append-system-prompt override per spawn (personas).
- "Always allow" stickiness for permissions (use SDK's `updatedPermissions`).
- Streaming or paginated tool results (currently truncated at 4 KB).
- Periodic cleanup of the SDK's on-disk session store.

## SDK quirks worth flagging upstream

1. **Native binary auto-detection picks the wrong variant** on some platforms (linux-x64-musl variant attempted on a glibc box where the musl optional dep wasn't installed). Worked around with explicit path resolution at startup.
2. **\`PermissionResult.updatedInput\` Zod-vs-TS mismatch** — TS type marks it optional, Zod validator requires a \`record\` on the allow branch. Returning \`{ behavior: "allow" }\` causes the SDK to reject the canUseTool reply with a \`ZodError\`. Worked around by echoing \`input\` back.

## Test plan

- [ ] \`bun run typecheck\` clean in \`examples/claude-code-headless\`
- [ ] \`bun run typecheck\` + \`bun run build\` clean in \`examples/agent-web-ui\`
- [ ] Bootstrap headless without NATS → graceful "connection refused" exit
- [ ] Bootstrap headless without \`ANTHROPIC_API_KEY\` → startup hint logged pointing at the README
- [ ] End-to-end via web UI:
  - [ ] CC Exec tab appears when a controller is discovered
  - [ ] Spawn a session with \`permission_mode: default\` + \`allowed_tools: Read,Glob,Grep\` succeeds
  - [ ] Prompt fires per-token text streaming + tool cards render inline
  - [ ] Permission prompt appears for a non-allowlisted tool (Write)
  - [ ] Allow click resolves canUseTool with \`updatedInput\`; tool actually executes
  - [ ] Cost annotation appears on agent bubbles + session card running total
  - [ ] Stop button cancels mid-turn cleanly
  - [ ] Multiple parallel sessions from one controller work independently
- [ ] PI Exec mode unaffected (regression check)
- [ ] Chat mode unaffected (regression check)

### DIFF
--- a/agents/claude-code/README.md
+++ b/agents/claude-code/README.md
@@ -228,6 +228,16 @@ NATS server authentication and authorization handle access control. If a
 user can connect and publish to `agents.cc.<user>.<name>`, they can
 interact with Claude. No additional pairing or allowlist is needed.
 
+## Anthropic auth
+
+Set `ANTHROPIC_API_KEY` in your environment before launching `claude`.
+Claude Code uses the env var in preference to any `~/.claude/`
+credentials, so logging out is unnecessary.
+
+Bedrock / Vertex / Azure deployments work too — set the standard
+provider env vars before launching `claude` and Claude Code will use
+those instead.
+
 ## Configuration
 
 State lives in `~/.claude/channels/nats/`:
--- a/examples/agent-web-ui/README.md
+++ b/examples/agent-web-ui/README.md
@@ -2,8 +2,9 @@
 
 A Bun + Vue 3 test client for the [`@synadia-ai/agents`](../../client-sdk/typescript) SDK.
 Discover agents over NATS, prompt them (with optional attachments),
-stream responses back, and — when a [`pi-headless`](../pi-headless) controller
-is online — spawn, prompt, and fan out PI sessions from the browser.
+stream responses back, and — when a [`pi-headless`](../pi-headless) or
+[`claude-code-headless`](../claude-code-headless) controller is online — spawn,
+prompt, and manage sessions from the browser.
 
 Primary use: manually poking at the [NATS Agent Protocol](https://github.com/synadia-ai/nats-agent-sdk-docs)
 implementations - [`pi`](../../agents/pi), [`claude-code`](../../agents/claude-code),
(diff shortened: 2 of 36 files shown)


M64GitHub:
Addressed all 5 review findings in `ebf3f28`:

- `pruneStale` now signals 408 timeout error before the terminator (no longer indistinguishable from successful completion)
- Attachment-staging error path uses two separate try-catch blocks (no caller hang if `respondError` throws)
- `toolUseStatus` bounds input size at 4 KB; preserves the record shape by swapping in a truncation marker (`{_truncated, _preview, _original_size_bytes}`) when oversized
- `generateSessionId` switched to `crypto.getRandomValues(Uint8Array(4))` for uniform 32-bit entropy (replaces biased `Math.random`)
- Config now validates `owner` / `name` as NATS-safe `[a-z0-9_-]{1,63}` tokens at boot, with a clear "override with --owner / --name" message

Plus, mirrored the entropy fix to `examples/pi-headless/src/subjects.ts` in `5dd4302` to keep the two headless examples consistent on session-id rng. Same fix, different file.

Please re-review.

M64GitHub:
@claude please re-review — all 5 prior findings addressed in `ebf3f28`, plus mirrored the entropy fix to pi-headless in `5dd4302`.

PR #33, opened 2026-04-29
client-sdks: bump default discover stall 200ms → 750ms (#31 — WAN broker fix)

### LINKED ISSUE #31: Discover default stall (200ms) too tight for WAN brokers — examples show "no agents found" against demo.nats.io
## Symptom

Running the bundled `02-prompt-text.py` (Python) or `02-prompt-text.ts` (TS)
against an agent registered on `demo.nats.io` consistently prints
`no agents found — start the reference agent first.`, even though the
agent is healthy and responding.

We've only directly tested `demo.nats.io`, but based on the root cause
below we expect any broker with RTT > ~200ms — likely including
Synadia Cloud from many client locations — to hit the same failure.

## Repro (SDK examples only)

One-time setup — register a `nats` CLI context for the demo broker:

```bash
nats context save demo-nats --server nats://demo.nats.io:4222
```

Window 1 — start the bundled reference agent against the demo broker:

```bash
uv run python client-sdk/python/examples/_reference_agent.py \
    --context demo-nats --session-name discover-stall-repro
```

Window 2 — run the canonical first-prompt example:

```bash
uv run python client-sdk/python/examples/02-prompt-text.py \
    --context demo-nats --session discover-stall-repro "hello"
# → "no agents found — start the reference agent first."
```

Verify the agent is actually live and responding on the same broker:

```bash
nats --context demo-nats req '$SRV.INFO.agents' '' --timeout 3s
# rtt ~315ms, full info_response returned
```

## Root cause

Both SDKs hardcode the same 200ms discover stall window:

- Python: `DEFAULT_DISCOVER_STALL_S = 0.2` — `client-sdk/python/src/synadia_ai/agents/discovery.py:68`
- TS:     `DEFAULT_DISCOVER_STALL_MS = 200` — `client-sdk/typescript/src/discovery/srv-ping.ts:18`

When broker RTT exceeds the stall window, `request_many_stall` times
out **before** the first `$SRV.INFO.agents` reply arrives and
`discover()` returns an empty list. `discover()` does expose `stall=`,
but neither bundled `02-prompt-*` example surfaces a flag, so a user's
first interaction with a WAN broker silently fails.

## Suggested fixes (any combination)

1. **Bump the default** to something that survives a transcontinental
   broker — e.g. 500–750ms. Cheap; "snappy on lightly-loaded systems"
   stays true at 500ms (still well under one perceptible UI tick).
2. **Expose `--stall` / `--timeout` on the example CLIs** (Python:
   `_connect_cli.py`; TS: equivalent loader). The examples are the
   canonical first-run surface — a flag is the lowest-friction escape
   hatch.
3. **Improve the empty-result message** in the examples: when
   `discover()` returns empty, hint that the broker may be far away
   and suggest `--stall 1.0` or `agents.discover(stall=1.0)`.

Local workaround: `agents.discover(stall=1.5)` — confirmed end-to-end
against demo.nats.io.

## Affected SDKs

Python and TypeScript — same constant, same example shape.

## Originally surfaced via

Hermes Agent's NATS gateway adapter, demoing on `demo.nats.io`. The
adapter is a thin wrapper around `AgentService`, so the failure
reproduces with the bundled reference agent above — Hermes is not in
the loop.

M64GitHub:
## Summary

Fixes [#31](https://github.com/synadia-ai/synadia-agents/issues/31). Both SDKs hardcode the same idle-window default for `discover()`'s stall strategy at **200 ms**. That's sized for a LAN broker. Against a WAN broker (issue surfaced this on `demo.nats.io`, ~315 ms RTT from a non-US client) `request_many_stall` times out *before* the first `\$SRV.INFO.agents` reply arrives, so `discover()` returns an empty list and the bundled `02-prompt-text.{ts,py}` example prints "no agents found" against a perfectly healthy agent.

Bumped the default to **750 ms** — comfortable margin over a transcontinental RTT (≥ 2× demo.nats.io's ~315 ms), still well under one perceptible UI tick on a LAN, still well under the existing 2 s `MAX_WAIT` safety cap. Callers that want a tighter window can still pass `timeoutMs` (timer strategy) or call the lower-level helper with their own `stall` / `stall_s`.

The change is wire-compat-safe — it's a client-side request-strategy constant, no protocol behaviour changes.

## What's in this PR (4 files)

| file | change |
|---|---|
| `client-sdk/typescript/src/discovery/srv-ping.ts` | `DEFAULT_DISCOVER_STALL_MS = 200 → 750` + docstring |
| `client-sdk/python/src/synadia_ai/agents/discovery.py` | `DEFAULT_DISCOVER_STALL_S = 0.2 → 0.75` + docstring |
| `client-sdk/typescript/CHANGELOG.md` | `[Unreleased]` entry |
| `client-sdk/python/CHANGELOG.md` | `[Unreleased]` entry |

Cross-SDK defaults are kept aligned per `client-sdk/python/CLAUDE.md`'s "wire compatibility" rule (when both SDKs pick the same default, drift is a bug).

## Out of scope (issue #31 also suggests)

The issue lists three fixes; this PR is suggestion **(1)** only:

1. ✅ **Bump the default** — this PR.
2. ⏳ **Expose `--stall` / `--timeout` on the example CLIs** — can be a separate PR per SDK; small scope, but UX-breaking if defaults change underneath.
3. ⏳ **Improve the empty-result message** in the examples — same.

Happy to do (2) and (3) in follow-up PRs once the bot review on this one is in.

## Verification

- **TS**: `bun run typecheck` clean. `bun run test:unit` — 201 / 201 passing.
- **Python**: `uv run ruff check --no-cache` clean on touched files. `uv run ruff format --check` clean. `uv run mypy --no-incremental` clean. `uv run pytest tests/test_discovery.py` — 4 / 4 passing.

## Note on TS `format:check`

The TS SDK's `Lint + typecheck` CI job has been failing on `main` for unrelated pre-existing whitespace drift across 6 files this PR doesn't touch (see GitHub Actions run `25077554126`). My CHANGELOG addition is itself prettier-clean — only the pre-existing drift remains. Cleaning that up is genuinely orthogonal to this fix; can ship as its own one-line PR.

### DIFF
--- a/client-sdk/python/CHANGELOG.md
+++ b/client-sdk/python/CHANGELOG.md
@@ -17,6 +17,15 @@ the 0.x line is explicitly unstable per protocol spec §11.2.
   (`_INBOX.agents.>`) covers caller-side reply traffic regardless of
   language. The connection's `inbox_prefix` is no longer consulted for
   agents-SDK reply subjects; not user-overridable.
+- `DEFAULT_DISCOVER_STALL_S` bumped from `0.2` → `0.75` so the default
+  `discover()` (stall strategy) survives a transcontinental NATS
+  round-trip — e.g. demo.nats.io reports ~315 ms RTT from a non-US
+  client, which previously caused `discover()` to return an empty
+  list before the first reply arrived. Snappy on LAN brokers stays
+  true at 750 ms (still well under one perceptible UI tick); callers
+  wanting a tighter window can pass `stall=` to `agents.discover()` /
+  `discover_agents()`. Fixes [#31]. Mirrors the same constant change
+  in the TypeScript SDK so cross-SDK defaults stay aligned.
 
 ### Changed (breaking, public API)
 
--- a/client-sdk/python/src/synadia_ai/agents/discovery.py
+++ b/client-sdk/python/src/synadia_ai/agents/discovery.py
@@ -65,7 +65,14 @@ STATUS_QUEUE_GROUP = "agents"
 
 # Idle window after the most recent reply before discover() returns
 # (stall strategy). Mirrors TS DEFAULT_DISCOVER_STALL_MS.
-DEFAULT_DISCOVER_STALL_S: float = 0.2
+#
+# Sized to comfortably absorb a transcontinental NATS round-trip — e.g.
+# demo.nats.io reports ~315 ms RTT from a non-US client. At 750 ms we
+# still return well under one perceptible UI tick on a LAN, but no
+# longer time out before the first reply arrives on a WAN (the symptom
+# that issue #31 surfaced). Callers who want a snappier scan can pass
+# `stall=` to `agents.discover()` or `discover_agents()`.
+DEFAULT_DISCOVER_STALL_S: float = 0.75
 
 # Absolute safety cap when using the stall strategy. Mirrors TS
 # DEFAULT_DISCOVER_MAX_WAIT_MS.
--- a/client-sdk/typescript/CHANGELOG.md
+++ b/client-sdk/typescript/CHANGELOG.md
@@ -112,6 +112,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
   tests that exercise it cover the new endpoint automatically. The
   `subjectAgentToken` option is gone — agents declare a single
   `agent` token used in both metadata and the subject.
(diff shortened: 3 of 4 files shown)


derekcollison:
The stall is about deltas between responses after the first response is received IIRC.

PR #41, opened 2026-04-30
max_payload: derive from nc.info, clamp on advertise, cap on publish

derekcollison:
Replace hard-coded 1MB max_payload across all agent harnesses and both SDKs with the broker-negotiated limit (nc.info.max_payload) read at connect time. Three layers, one rule: advertise what the broker will accept; never advertise more than the broker allows; never publish more than the smaller of the agent's advertised cap and the caller's own broker cap.

  Agent harnesses (TS):
    - agents/{claude-code,openclaw,pi}: read nc.info.max_payload at startup, format back to the §2.1 \d+(B|KB|MB|GB) grammar via a new formatMaxPayloadString helper, advertise that value on the prompt endpoint metadata, and use it for §5.4 inbound enforcement and response-chunk splitting. DEFAULT_MAX_PAYLOAD_{STR,BYTES} = "1MB" / 1MiB are the fallbacks used only when nc.info.max_payload is absent.
    - Naming consistent across the three harnesses (was a mix of MAX_PAYLOAD_STR_FALLBACK / MAX_PAYLOAD_STRING_FALLBACK).
    - Smoke tests now spawn their own private nats-server on an ephemeral port (port: -1 + --ports_file_dir, configured via test/nats-server.conf with max_payload: 8MB) so the dynamic path is exercised every run instead of accidentally matching 1MB. Never touches a developer's server on 4222. PI smoke writes a throwaway NATS context so the harness's NATS_CONTEXT > config > NATS_URL chain lands on the test server even when ~/.pi/agent/nats-channel.json points elsewhere.

  Server-side SDKs (AgentService):
    - Python (synadia_ai.agents.service) and TS (@synadia-ai/agents, plus the testing ReferenceAgent) now clamp a constructor-supplied max_payload override down to nc.{info.,}max_payload at start() and log a warning. Smaller overrides are honored (use case: shed expensive prompts before the handler). DEFAULT_MAX_PAYLOAD = "1MB" is the fallback when the server didn't report a value.
    - Adds {format_human_bytes,formatHumanBytes} as the inverse of the existing parsers (largest-clean-unit; "8MB" round-trips, not "8192KB").

  Caller-side SDKs (Agent.prompt):
    - assert_within_max_payload / assertWithinMaxPayload gain an optional connection_max_payload / connectionMaxPayload param. Effective cap is min(endpoint.max_payload_bytes, nc.{info?.,}max_payload) when both are known; the smaller of which is set otherwise; no check when neither is. A caller behind a smaller-cap broker (multi-cluster / per-account) now fails fast with PayloadTooLargeError instead of waiting for the broker's MAX_PAYLOAD_VIOLATION rejection.

  Tests:
    - Python: 211 passed (was 202) — bytes round-trip, AgentService clamp matrix (5 cases), validation connection-limit matrix (4 cases).
    - TS: 287 passed (was 283) — same matrices in unit tests, plus a real-broker integration test that constructs AgentService with maxPayload="16MB" against the test server's 1MB and asserts the advertised metadata is "1MB" + console.warn captured.
    - Smoke tests (openclaw, pi) verified end-to-end against an 8MB ephemeral server.

  Docs:
    - README.md, client-sdk/README.md, client-sdk/typescript/README.md, agents/README.md, agents/{claude-code,openclaw,pi,hermes}/README.md, client-sdk/{python,typescript}/docs/protocol-mapping.md updated to describe the server-derived value, the clamp behavior, and the min(endpoint, nc) caller-side rule.
    - CHANGELOG entries in both SDKs under [Unreleased].

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ Full spec: <https://github.com/synadia-ai/nats-agent-sdk-docs>
        └─── streamed chunks ───────┘
 ```
 
-- **`client-sdk/*`** - produce envelopes, validate locally against agent metadata (`max_payload`, `attachments_ok`), parse streamed chunks.
+- **`client-sdk/*`** - produce envelopes, validate locally against agent metadata (`max_payload`, `attachments_ok`) and the caller's own `nc.info.max_payload` (the smaller of the two binds — the caller's broker rejects oversized publishes before they reach the agent), parse streamed chunks.
 - **`agents/*`** - register the `agents` micro service, drive the underlying AI harness, stream chunks back.
 - **`examples/*`** - demonstrate SDK usage end-to-end against real agents.
 
--- a/agents/README.md
+++ b/agents/README.md
@@ -8,10 +8,12 @@ For an example of *building* a fresh agent from scratch using the TypeScript SDK
 
 | Path                | Type token | Underlying harness                          | Prompt subject (v0.3 verb-first)                              | `max_payload` | `attachments_ok` |
 | ------------------- | ---------- | ------------------------------------------- | ------------------------------------------------------------- | ------------- | ---------------- |
-| `pi/`               | `pi`       | [PI Agent](https://github.com/badlogic/pi-mono) | `agents.prompt.pi.<owner>.<session>`                      | 1 MB          | true             |
-| `openclaw/`         | `oc`       | [OpenClaw](https://openclaw.ai)             | `agents.prompt.oc.<owner>.<agentName>`                        | 1 MB          | true             |
-| `claude-code/`      | `cc`       | [Claude Code](https://claude.com/claude-code) | `agents.prompt.cc.<owner>.<name>`                          | 1 MB          | true             |
-| `hermes/`           | `hermes`   | [Hermes Agent](https://github.com/NousResearch/hermes-agent) | `agents.prompt.hermes.<owner>.<name>`          | 1 MB (configurable) | true        |
+| `pi/`               | `pi`       | [PI Agent](https://github.com/badlogic/pi-mono) | `agents.prompt.pi.<owner>.<session>`                      | 1 MB / system-defined | true |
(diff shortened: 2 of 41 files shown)


M64GitHub (commented):
Three compounding errors:
  1. const declaration as process.stderr.write(...) argument — parser refuses to load the file.
  2. Two const maxPayload in the same block.
  3. Reference to MAX_PAYLOAD_BYTES (the old constant the same PR renamed to DEFAULT_MAX_PAYLOAD_BYTES).
  4. The original log message nats channel: replying to ${pending.replySubject}... was lost as
  collateral.

  Why it slipped through CI: agents/claude-code/ has no typecheck script (only bun server.ts at
  runtime), and PR #41 didn't add a smoke test for claude-code (only openclaw + pi got test/smoke.mjs
  updates). Bun would refuse to start the agent at all.

  Fix: restore the process.stderr.write(\nats channel: replying to ${pending.replySubject}...`,)line and
   drop the strayconst maxPayload = MAX_PAYLOAD_BYTESinsertion. The legitimateconst maxPayload =
  nc.info?.max_payload ?? DEFAULT_MAX_PAYLOAD_BYTES` below it should stay.

  Blocker 2: agents/pi/test/smoke.mjs references undefined symbols

  Lines 213–214 use formatMaxPayloadString and DEFAULT_MAX_PAYLOAD_STR, but the file never imports them.
   And agents/pi/extensions/nats-channel.ts declares both as module-private (no export), so adding an
  import wouldn't even work.

  Compare to openclaw, which exports both from src/nats/protocol.ts and the smoke imports them correctly
   — that side is fine.

  Also: a comment block is duplicated above the assertion (cosmetic).

  Fix: add export to formatMaxPayloadString and DEFAULT_MAX_PAYLOAD_STR in
  agents/pi/extensions/nats-channel.ts, then import them in smoke.mjs. Remove the duplicated comment.

  Everything else: ✅

M64GitHub:
@claude - please review again

M64GitHub:
@claude please review again

M64GitHub:
@claude all resolved, please check again

PR #48, opened 2026-05-01
Phase 3: split TypeScript SDK into caller (@synadia-ai/agents) + host (@synadia-ai/agent-service)

M64GitHub:
## Summary

Splits `@synadia-ai/agents` (TypeScript SDK at 0.4.0) into two npm
packages, mirroring the Python split landed in PR #45 and the layout
already advertised by the Hermes README in PR #47.

- **`@synadia-ai/agents`** (caller, slim) — discovery, prompt, heartbeat
  tracking, wire codecs, NATS context loading. Most consumers (e.g.
  `examples/agent-web-ui`) want only this. Same package name; existing
  caller-only consumers need **zero** code changes.
- **`@synadia-ai/agent-service`** (host, NEW) — `AgentService`,
  `ReferenceAgent`, server-side wire helpers. Lives at the new
  `agent-sdk/typescript/` sibling directory. Imports
  `@synadia-ai/agents` as a regular runtime dep.

Both packages stay at version **0.4.0** in lockstep. Neither has been
published to npm — the dev-cycle uses `file:` links between sibling
dirs. First publish is gated on user approval and out of scope here.

The host's `AgentService` gains two new extension points so a future
follow-up can migrate `examples/pi-headless` / `examples/claude-code-headless`
controllers off raw `Svcm`:

- `AgentServiceOptions.extraEndpoints` — declarative custom endpoints
  registered alongside `prompt` and `status` (e.g. `spawn`/`stop`/`list`).
  Names validated for collisions before any registration.
- `AgentService.service` getter — escape hatch returning the underlying
  `@nats-io/services` `Service` for runtime-dynamic endpoints.

Plan source:
[`devtools/plans/sdk-split-typescript.md`](https://github.com/synadia-ai/synadia-agents-devtools/blob/main/plans/sdk-split-typescript.md)
(self-contained — read top to bottom). Locked decisions and the
exhaustive move boundary are documented there.

## Commits

**The 9 plan phases:**

1. `client-sdk/typescript: scaffold sibling agent-sdk/typescript/ package` (ca7d771)
2. `agent-sdk/typescript: move host-side source via git mv` (660851b) — `git mv` preserves blame; tree intentionally doesn't typecheck mid-phase.
3. `agent-sdk/typescript: rewrite moved-file imports to @synadia-ai/agents` (b9ea170)
4. `client-sdk/typescript: trim caller index.ts and CHANGELOG for split` (cdeb200) — caller fully self-consistent again.
[...]

PR #49, opened 2026-05-01
examples: add `bin` to pi-headless + claude-code-headless so they run via npx

M64GitHub:
## Summary

The first publish of `@synadia-ai/nats-pi-headless@0.4.0` shipped TS source with a
`bun run src/index.ts` start script. Reachable from a local clone, but no path to
run the service from npm without cloning. Same gap on the parallel
`claude-code-headless` example (which was queued for first publish).

This PR mirrors the `@m64/nats-pi-bridge` shape: bundled `dist/index.js` with a
shebang + a `bin` entry, so:

```sh
npx @synadia-ai/nats-pi-headless --context localhost
npx @synadia-ai/nats-claude-code-headless --context localhost
```

…Just Works. Both packages bump to **0.4.1**.

## What changed in each example

- `bin: { "<name>": "dist/index.js" }` — two new CLI commands.
- `engines: { node: ">=20" }`.
- `build` = `esbuild src/index.ts --bundle --platform=node --format=esm --packages=external --banner:js='#!/usr/bin/env node' --outfile=dist/index.js && chmod +x dist/index.js`.
- `prepack` = `npm run build` so `npm publish` always ships a current `dist/`.
- `files` trims to `["dist", "README.md", "LICENSE"]` — TS source stays in the repo for clone consumers but no longer in the npm tarball.
- `esbuild ^0.25.0` added as devDep.
- `start: "bun run src/index.ts"` kept so the local-clone dev workflow is unchanged.

READMEs lead with a "Quickstart (run from npm)" section; the pre-existing
local-clone instructions move under "Quickstart (run from a local clone)" and
keep the build-both-SDKs dance for SDK contributors.

## Test plan

- [x] `bun run build` in both packages — 30.8 kB pi-headless / 44.7 kB cc-headless, shebang + `+x` confirmed.
- [x] `./dist/index.js` runs directly under Node and exits with the expected "no NATS target" error when no context/URL is set.
- [x] `npm pack --dry-run` fires the `prepack` hook automatically, ships 4 files per package.
- [x] `bun run typecheck` green for both.
- [ ] Post-merge: re-publish `@synadia-ai/nats-pi-headless@0.4.1` and first-publish `@synadia-ai/nats-claude-code-headless@0.4.1` via the standard release ladder.
- [ ] End-to-end: `npx @synadia-ai/nats-pi-headless --context localhost` connects to a real broker and registers the controller (post-publish).

### DIFF
--- a/examples/claude-code-headless/README.md
+++ b/examples/claude-code-headless/README.md
@@ -10,23 +10,54 @@ This example is the Claude Code analogue of [`examples/pi-headless/`](../pi-head
 
 > **Status.** Demo-quality reference. Streams text per-token, surfaces tool calls + results inline, asks for permission via protocol §7 query chunks, and tracks cost per session. See **Features** below for the wire shapes; **Roadmap** for what's still ahead.
 
-## Quickstart
+## Quickstart (run from npm)
+
+The package ships a `nats-claude-code-headless` CLI binary, so the
+simplest way to try it is via `npx` — no clone, no build:
 
 ```bash
-# 1. Build both SDKs (workspace siblings, referenced via file:). The
-#    extra `bun install` in agent-sdk re-copies the freshly-built
-#    caller dist into agent-sdk/node_modules/@synadia-ai/agents/, which
-#    is the path the host SDK's compiled output resolves at runtime.
+export ANTHROPIC_API_KEY=sk-...
+NATS_CONTEXT=localhost npx @synadia-ai/nats-claude-code-headless
+# or:
+NATS_URL=nats://127.0.0.1:4222 npx @synadia-ai/nats-claude-code-headless
+# or:
+npx @synadia-ai/nats-claude-code-headless --context localhost
+```
+
+`npx` resolves the package, runs its bundled entry point under Node ≥ 20,
+and prints:
+
+```
+claude-code-headless: controller listening on agents.prompt.cc.<you>.exec
+claude-code-headless: extra endpoints — …exec.spawn  …exec.stop  …exec.list
+```
+
+For a permanent install:
+
+```bash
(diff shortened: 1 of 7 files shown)


M64GitHub:
The reviewer's blocking bug doesn't actually fire — its premise (the SDKs aren't published to npm) is stale. `@synadia-ai/agents@0.4.0` and `@synadia-ai/agent-service@0.4.0` went up earlier today (PR #48 release ladder). Confirmed:

```sh
$ npm view @synadia-ai/agents@0.4.0 version
0.4.0
$ npm view @synadia-ai/agent-service@0.4.0 version
0.4.0
```

Ran the would-be `0.4.1` end-to-end against a clean install dir to verify:

```sh
# devmode.sh off → file: → ^0.4.0 across the consumers, then:
$ npm pack
synadia-ai-nats-pi-headless-0.4.1.tgz

$ mkdir /tmp/scratch && cd /tmp/scratch && npm init -y
$ npm install /path/to/synadia-ai-nats-pi-headless-0.4.1.tgz
added 36 packages in 4s, found 0 vulnerabilities

$ ./node_modules/.bin/nats-pi-headless
pi-headless: fatal: pi-headless: no NATS target configured.
```

That last line is exactly the right error path — Node loaded the bundle, resolved every `@synadia-ai/*` and `@nats-io/*` import from the clean `node_modules`, and aborted on missing config (not on missing deps). The `--packages=external` shape is intentional and matches the reference (`@m64/nats-pi-bridge`).

Bundling `@synadia-ai/*` into the binary would work too but has a real downside: SDK bug fixes wouldn't reach `nats-pi-headless` consumers without re-publishing the example. With externals + caret semvers, a `0.4.x` SDK patch flows through `npm update` automatically.

Keeping the current shape; merging.

PR #50, opened 2026-05-01
devtools/: relocate devmode.sh into the repo + document the release ladder

M64GitHub:
## Summary

Moves the dev/release toggle script (`devmode.sh`) from a sibling out-of-tree
checkout into this repo at `devtools/`, and documents the full release ladder
in `README-DEV.md`. After this lands, any maintainer with publish rights can
drive an SDK + consumer release from a fresh clone — no private tooling repo
required.

## Files

- **`devtools/devmode.sh`** — the script. Two changes vs. the out-of-tree copy:
  - Resolves the repo as `$HERE/..` (was `$HERE/../synadia-agents`). The
    `REPO=/path` override still works for unusual layouts.
  - Wraps each per-consumer `bun install` in `timeout 60` (configurable via
    `BUN_INSTALL_TIMEOUT`). `agents/openclaw`'s `"openclaw": ""` peer dep
    sends bun into a 100%-CPU walk that never returns — we hit it twice
    during the 0.4.x rollout. The wrapper prints a `⏱ timed out` line and
    continues; the package.json flip itself always lands regardless of the
    lockfile-refresh outcome.
- **`devtools/.devmodeignore`** — verbatim copy. Lists `dspy` (private example
  that stays on `file:` permanently).
- **`devtools/README.md`** — short overview pointing at the script and the
  ladder doc.
- **`README-DEV.md`** — new "Releasing the SDKs" section walking one full
  publish cycle (off → publish caller → publish host → publish consumers →
  on), plus two new troubleshooting rows for the openclaw spin and the
  pre-publish `^semver` resolve errors.

## Test plan

- [x] `./devtools/devmode.sh status` from repo root — discovers 13 SDK-dep refs across `agents/*`, `examples/*`, `agent-sdk/*`, `client-sdk/*`, skipping `dspy`.
- [x] `./devtools/devmode.sh status` from `/tmp` with `REPO=…` — works.
- [x] Script syntax check (`bash -n`) — clean.
- [x] `bash -n` clean; help text updated.
- [ ] Post-merge: try `./devtools/devmode.sh off` → wait for the openclaw timeout (≤60 s) → `./devtools/devmode.sh on` to confirm the new timeout path works end-to-end against `agents/openclaw`.

## Out of scope

- The out-of-tree copy at `~/space/synadia-ai/synadia-agents-devtools/` isn't
  modified. Whether to delete it or leave a redirect pointer is the
  maintainer's call.
- The `openclaw-reinstall.sh` script in that sibling repo stays where it is —
  it's deployment-specific (docker-container redeploy for a particular
  operator's openclaw farm), not a generic release tool.

### DIFF
--- a/README-DEV.md
+++ b/README-DEV.md
@@ -134,6 +134,87 @@ CI runs the same shape — see
 [`.github/workflows/client-sdk-typescript.yml`](.github/workflows/client-sdk-typescript.yml)
 and [`.github/workflows/agent-sdk-typescript.yml`](.github/workflows/agent-sdk-typescript.yml).
 
+## Releasing the SDKs
+
+`main` keeps `file:` links between consumers and the SDK packages so
+contributors editing the SDK see their changes live in the
+agents/examples without any flip step. That's also why a fresh `npm
+publish` of any consumer would ship `file:` refs that break for npm
+users — published tarballs need `^semver` instead. The
+[`devtools/devmode.sh`](devtools/devmode.sh) script bridges the two
+states.
+
+```sh
+./devtools/devmode.sh status        # what's currently flipped where
+./devtools/devmode.sh off           # flip every tracked consumer to ^semver
+./devtools/devmode.sh on            # flip back to file: (the default state)
+./devtools/devmode.sh check-release # exit 0 iff every dep is at its SDK's ^semver
+```
+
+The script discovers consumers automatically — every `package.json`
+under `examples/`, `agents/`, `client-sdk/`, and `agent-sdk/` that
+depends on a tracked SDK gets flipped. Names listed in
+`devtools/.devmodeignore` are skipped (currently just `dspy`, which
+lives on `file:` permanently).
+
+### The release ladder (one cycle)
+
+Order matters: caller `@synadia-ai/agents` first because the host SDK
+declares `^0.4.x` against it; agent harnesses and headless examples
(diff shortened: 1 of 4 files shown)


M64GitHub:
Addressed all four findings in 13f00de:

1. **Hardcoded `file:../../`** → new `file_ref_for($consumer, $sdk_idx)` helper counts slashes in the consumer path and emits the right number of `../`s. Output is byte-identical for every current depth-2 consumer (verified via `status`); deeper future consumers would now produce correct refs instead of silently broken ones.
2. **`BUN_INSTALL_TIMEOUT` validated** at preflight with a `^[0-9]+$` regex. `BUN_INSTALL_TIMEOUT=abc` now exits 64 with a clear config error before any work, instead of cascading as a per-consumer install failure.
3. **`set_ref` cleans up `.tmp`** on either jq or mv failure — explicit `if jq … && mv …; then return 0; fi` with `rm -f` on the failure path and an `err` message. No more orphaned `.tmp` files when a write fails midway.
4. **`mapfile` → portable while-read** in both call sites (IGNORED + CONSUMERS). Tested fine on bash 5.2 here; the new shape is bash 3.2-compatible per the macOS portability concern.

`./devtools/devmode.sh status` round-trips clean, dynamic refs match the previous static ones byte-for-byte.

PR #55, opened 2026-05-01
ci(release): grant contents:write so GH Release attach step doesn't 403

renerocksai:
## Summary

Two-line fix in both Python release workflows:
`contents: read` → `contents: write` on the `publish` job's
`permissions` block.

`softprops/action-gh-release@v2`'s "create release" call hits the
GitHub REST releases API, which requires write scope on `contents`.
With `contents: read`, the call returns `403 Resource not accessible
by integration` and the step fails after PyPI publishing has already
completed.

### Observed on `python-v0.5.0`

The first `synadia-ai-agents` publish ran the release ladder
end-to-end:

- ✅ `verify` matrix (py3.11/3.12/3.13) green
- ✅ `Build sdist + wheel`
- ✅ `Verify tag matches pyproject version`
- ✅ `Publish to PyPI` (OIDC trusted publishing — package is live at
  https://pypi.org/project/synadia-ai-agents/0.5.0/)
- ❌ `Attach artifacts to GitHub release` — `403 Resource not
  accessible by integration`

Failed run: https://github.com/synadia-ai/synadia-agents/actions/runs/25228697053

The PyPI side is fine; this fix unblocks the post-publish GH Release
creation for the **next** publish.

### Same fix applied to both files

The bug is identical in `release-python-agent-service.yml`. Without
this fix, the agent-sdk's first publish (next on the ladder) would
hit the same 403 mid-workflow.

### Why not narrow `contents:write` to a separate job?

Single-job design keeps the OIDC exchange + GH-Release-attach in one
runner with one checkout. Splitting them adds a job for marginal
permission narrowing on a workflow that only fires on tag pushes
gated by the `pypi` GitHub Environment.

`id-token: write` is unchanged (still required for OIDC publish).

## Test plan

- [ ] Reviewer-bot pass on the workflow diff.
- [ ] Out-of-band: backfill `python-v0.5.0`'s GH Release with `gh
      release create python-v0.5.0 --generate-notes` after merge
      (separate from this PR — the `0.5.0` tag has already published
      to PyPI; re-running the workflow on the same tag would attempt
      a duplicate PyPI upload and fail).
- [ ] Live test: agent-sdk's first `python-agent-service-v*` tag
      should publish to PyPI **and** create a GitHub Release in one
      run.

### DIFF
--- a/.github/workflows/release-python-agent-service.yml
+++ b/.github/workflows/release-python-agent-service.yml
@@ -104,7 +104,11 @@ jobs:
           packages-dir: agent-sdk/python/dist/
 
       - name: Attach artifacts to GitHub release
-        if: ${{ !cancelled() }}
+        # success() (not !cancelled()) so a failed PyPI publish doesn't
+        # leave behind a misleading GitHub Release pointing at a version
+        # that doesn't exist on PyPI — uv build still produced the
+        # wheel/sdist locally, so the dist/ glob would match either way.
+        if: success()
         uses: softprops/action-gh-release@v2
         # `files:` globs resolve from $GITHUB_WORKSPACE (defaults.run.working-directory
         # does not apply to action inputs), so paths are given from the repo root.
--- a/.github/workflows/release-python-agent-service.yml
+++ b/.github/workflows/release-python-agent-service.yml
@@ -71,7 +71,7 @@ jobs:
     # release workflow.
     environment: pypi
     permissions:
-      contents: read
+      contents: write  # softprops/action-gh-release creates the tag's GitHub Release
       id-token: write  # required for PyPI trusted publishing (OIDC)
     steps:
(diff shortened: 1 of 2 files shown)


renerocksai:
Good catch on the `if: ${{ !cancelled() }}` — pushed `2021f55` to fold the tightening into this PR.

One small correction to the framing: the `dist/` glob *would* match files even on a failed publish, because `uv build` runs in an earlier step and writes the wheel + sdist to disk independently of whether PyPI accepted them. So the failure mode wasn't "GH Release with no artifacts" — it was "GH Release with artifacts attached, pointing at a PyPI version that doesn't exist." Worse, not better. Result is the same fix: `if: success()`.

Comment in both workflows explains the why.

@claude, please review again.

PR #57, opened 2026-05-01
docs(agent-sdk): correct extraction lineage in [0.1.0] (was 0.5.0, should be pre-0.5.0)

renerocksai:
## Summary

Two refs in `agent-sdk/python/CHANGELOG.md`'s `[0.1.0]` entry framed
the agent-host surface as "extracted from `synadia-ai-agents@0.5.0`."
That gets the relationship backward — `0.5.0` is the version that
**removed** the surface from `synadia-ai-agents`, not the version
that contained it. The surface lived in `synadia-ai-agents` through
`0.4.x`; both packages were cut together at the 0.5.0 / 0.1.0 split.

Worse, the `AgentService` bullet originally said *"Sourced from
`synadia-ai-agents@0.5.0`'s `service.py`"* — `synadia-ai-agents@0.5.0`
doesn't have a `service.py` at that path. The file moved here. So
the original text was literally false on a path you can grep.

This PR catches that wording before it ships inside the
`synadia-ai-agent-service@0.1.0` wheel — re-tagging after publish
would mean a `0.1.1` cut just to fix CHANGELOG framing.

### Changes

- `[0.1.0]` preamble now reads "Carved out of `synadia-ai-agents`
  at the 0.5.0 cut" and clarifies the 0.4.x / 0.5.0 split.
- `AgentService` bullet now reads "Sourced from `synadia-ai-agents`'s
  pre-0.5.0 `service.py` (the file moved here at the split)."

No code change. Doc-only.

## Test plan

- [ ] Reviewer-bot pass on the doc diff.
- [ ] After merge: tag `python-agent-service-v0.1.0` with the
      already-corrected tag-annotation message.

### DIFF
--- a/agent-sdk/python/CHANGELOG.md
+++ b/agent-sdk/python/CHANGELOG.md
@@ -8,19 +8,24 @@ the 0.x line is explicitly unstable per protocol spec §11.2.
 
 ## [0.1.0] - 2026-04-30
 
-Initial release. **Extracted from `synadia-ai-agents@0.5.0`** so the
-agent-host surface ships as a focused dependency for harness authors
-while callers can install just the client SDK.
+Initial release. **Carved out of `synadia-ai-agents` at the 0.5.0
+cut**: through 0.4.x the agent-host surface lived inside
+`synadia-ai-agents`; the 0.5.0 release removed it there and shipped
+it here as 0.1.0. Both packages were cut together —
+`synadia-ai-agents@0.5.0` is the first PyPI version that no longer
+carries this surface. Harness authors get a focused dependency;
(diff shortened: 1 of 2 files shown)


renerocksai:
Good catch on `CLAUDE.md` — pushed `a6cfeb6` to fold the same fix in. Both spots (Status block, Alignment milestones) now read "carved out of \`synadia-ai-agents\` at the 0.5.0 cut" with the 0.4.x lineage made explicit, matching the CHANGELOG framing. Verified `grep extracted/Extracted` in `agent-sdk/python/CLAUDE.md` returns no hits.

@claude, please review again.

PR #58, opened 2026-05-01
docs(release): note the pypi env tag-policy gotcha + list both Python releases

renerocksai:
## Summary

Updates the Python release-ladder paragraph in root `CLAUDE.md` to:

1. **List both Python release workflows / tag prefixes** —
   `python-v*` for `synadia-ai-agents` and `python-agent-service-v*`
   for `synadia-ai-agent-service`. The previous text was written
   when only the first existed and read as if the env was
   single-purpose.
2. **Document the `pypi` env tag-policy gotcha** that bit the
   `synadia-ai-agent-service@0.1.0` publish: env protection
   rejected the tag because only `python-v*` was in the
   deployment-branch policy at the time. The workflow YAML was
   correct; the gap was on the GitHub-side env config.
3. **Provide the exact `gh api` one-liner** to extend the env
   policy when adding a new tag-triggered release workflow, plus
   the admin-check command to verify perms.

### Why now

This sharp edge cost ~15 min and a chain of *"can you grant
admin? wait, did the role propagate? wait, was Phil joking?"*
turn-arounds during the agent-sdk first publish. Future
contributors adding a third release workflow (e.g. when the TS
SDK gets a tag-driven publish) will hit the same wall unless the
ladder section warns them.

### Out of scope

- Updating `client-sdk/python/CLAUDE.md` or `agent-sdk/python/CLAUDE.md`
  with the same gotcha. The package-level files defer to root for
  release-ladder details, so a single mention here is enough.

## Test plan

- [ ] Reviewer-bot pass on the doc diff.
- [ ] Spot-check that the `gh api` POST and admin-check commands
      copy-paste cleanly (no smart-quote mangling from this PR
      description).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -208,9 +208,12 @@ gh api -X POST repos/synadia-ai/synadia-agents/environments/pypi/deployment-bran
   -f name='new-prefix-v*' -f type=tag
 ```
 
-(Repo Admin required —
+(Repo Admin required. Sanity check:
 `gh api repos/synadia-ai/synadia-agents --jq '.permissions.admin'`
-should return `true`.)
+should print exactly `true`. Anything else means no admin in the
+current auth context — `false` is a real "you don't have it,"
+while `null` typically means a fine-grained PAT or GitHub App
+token that doesn't surface the permissions sub-object at all.)
 
 ## CI and the Claude reviewer bot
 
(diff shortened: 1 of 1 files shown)


renerocksai:
Good call on `null` vs `false` — pushed `b3f5fef` to spell out all three cases (`true` / `false` / `null`) so a future reader hitting `null` under a fine-grained PAT doesn't mistake it for a missing admin grant. Thanks!

@claude, please review again.

PR #59, opened 2026-05-01
docs(hermes): drop pre-PyPI install block; PyPI is the default path

renerocksai:
## Summary

- `synadia-ai-agents 0.5.0` and `synadia-ai-agent-service 0.1.0` are live on PyPI, so `agents/hermes/README.md` no longer needs the "Pre-PyPI install — required until both ship to PyPI" `<details>` block.
- Replace it with a short **Monorepo development** `<details>` that explains when sibling-checkout via `[tool.uv.sources]` is useful (co-iterating on the SDKs and Hermes) and what plain PyPI users get.
- Inline comment in `setup-hermes.sh` block now correctly says the `[nats]` extra pulls both SDKs from PyPI.
- Troubleshooting entry for the "SDKs not installed" log line points at the new section instead of the deleted one.
- Attachment-example comment that referenced the old "Directory-layout section" now points to "Monorepo development" and notes that any local image works for non-sibling checkouts.

## Follow-up (separate repo)

The hermes-agent fork's `pyproject.toml` still carries `[tool.uv.sources]` redirecting both SDKs to `../synadia-agents/`. That cleanup lands in a separate PR on `synadia-ai/hermes-agent` (branch `nats-gateway`); without it, `uv sync` continues to prefer sibling-checkout when present, which is the behavior the new "Monorepo development" block describes.

## Test plan

- [x] PyPI versions verified (`synadia-ai-agents 0.5.0`, `synadia-ai-agent-service 0.1.0`).
- [x] No remaining `pre-PyPI` / "not yet" references in the README outside the still-accurate fork-vs-upstream banner.
- [ ] Reviewer skim of the rendered `<details>` block on the PR page (rendering checks).

### DIFF
--- a/agents/hermes/README.md
+++ b/agents/hermes/README.md
@@ -32,13 +32,6 @@ cd hermes-agent
 
 After this you can run `hermes --help` from anywhere. User state lives in `~/.hermes/`.
 
-The `[nats]` extra resolves two PyPI distributions:
-
-- **`synadia-ai-agents`** (≥ 0.5) — wire-types client SDK (import root `synadia_ai.agents`).
-- **`synadia-ai-agent-service`** (≥ 0.1) — host-side agent SDK (import root `synadia_ai.agent_service`, provides `AgentService` / `PromptStream`).
-
-Without these, Hermes logs `NATS: synadia-ai-agents / synadia-ai-agent-service SDKs not installed` at startup and skips registering the NATS adapter.
-
 ### 2. Configure the gateway
 
 Edit `~/.hermes/config.yaml` and add the `platforms.nats` block. The `owner` and `session_name` fields determine your subject — `agents.prompt.hermes.<owner>.<session_name>`.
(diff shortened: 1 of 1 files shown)


renerocksai:
Thanks for the catch on line 57 — you were right, the "PyPI fallback" sentence described a state that doesn't exist yet (the `[tool.uv.sources]` block is still live in the fork). Rather than caveat it with a "until the fork PR lands…" note, I went further: removed the entire **Monorepo development** `<details>` block, and the orphan paragraph that was its lead-in.

Reasoning: this README is the user-facing install guide for *running* a NATS-enabled Hermes. Monorepo-dev workflow (sibling checkout, `[tool.uv.sources]`, editable installs) is maintainer-side concern that belongs with whoever owns the fork, not in the default-case quickstart. Dragging the reader through it was a category error on my part — the original `<details>` was scaffolding for the pre-PyPI transition, and once both packages shipped to PyPI it had no audience left.

Cross-references that pointed at the deleted block (attachment-example comment, troubleshooting entry) are simplified accordingly. The fork's `[tool.uv.sources]` cleanup lands in a separate PR on `synadia-ai/hermes-agent`.

@claude, please review again.

PR #64, opened 2026-05-02
docs(hermes): document setup-wizard NATS flow + consolidate Configure

renerocksai:
## Summary

- Hermes's setup wizard now offers NATS as a first-class platform alongside Telegram/Discord/Slack. Rewrite `agents/hermes/README.md` `## Install` as a single wizard-led narrative: clone + bootstrap → wizard prompt → platform multi-select (screenshot) → NATS connection picker (screenshot) → owner/session prompts → confirmation → service-or-foreground start (with the actual banner + harmless allowlist warning users see).
- Fold the YAML quickstart that was duplicated between old Install §2 and `## Configure` into a single Configure section. Lead-in explains who the section is for; old YAML quickstart moved (not removed) into a new `### Minimal demo.nats.io setup`. CLI-context recipe, config-field table, and env-var table all preserved unchanged.
- New `agents/hermes/images/` with two screenshots (`setup-wizard-platforms.png`, `setup-wizard-nats.png`).

Wizard prompt strings cross-checked against the upstream sources:
- `setup-hermes.sh:393` — post-install Y/n prompt
- `hermes_cli/setup.py:1769–1810, 1856` — NATS connection menu, owner/session, success line
- `hermes_cli/setup.py:2538` — service-install offer
- `hermes_cli/gateway.py:1989–1994` — startup banner
- `gateway/run.py:1960–1963` — allowlist warning

## Test plan

- [ ] Render the README in GitHub preview mode; confirm both screenshots load and the section ordering reads naturally end-to-end.
- [ ] Spot-check that wizard text matches the source strings verbatim (no paraphrasing of literal on-screen output).
- [ ] Walk `setup-hermes.sh` end-to-end against a fresh `HERMES_HOME` and verify the documented flow matches reality.
- [ ] `git diff main` touches only `agents/hermes/README.md` + the two new image files.


renerocksai:
Thanks for the review. Fixup in cb13c7a addresses both findings:

1. **Anchor on line 47** — now points at `#via-a-nats-cli-context-recommended-for-anything-beyond-demonatsio` rather than the parent `#configure`.
2. **Foreground visibility of `[Nats] Connected — …`** — went a different direction than the suggested Configure-side hint. The underlying problem is that the line is INFO-level and `hermes gateway run` defaults to WARNING, so even users following the manual Configure path can't see it without `-v`. Sweeping the two Troubleshooting bullets that already reference that log message with a `-v`-or-journal hint felt cleaner than restoring an "expect:" comment in Configure that would silently fail to materialize for foreground users. The `(server-negotiated)` vs `(configured)` `max_payload` semantics are still documented in `## How it works` point 3.

@claude, please review again.

PR #66, opened 2026-05-03
Feature(client-sdk/ts): Drive PromptStream with requestMany + sentinel

derekcollison:
## Summary

  - Switch `PromptStream` to `nc.requestMany(subject, payload, { strategy: "sentinel", maxWait })`. The §6.5 terminator's empty body is the sentinel; the response routes
  through the connection's mux inbox instead of a per-stream `_INBOX.agents.>` subscription. Wire shape unchanged.
  - Add `PromptOptions.maxWaitMs` (override) and `DEFAULT_PROMPT_MAX_WAIT_MS = 600_000` (10 min default). Independent of the existing `inactivityTimeoutMs` — one is an
  absolute ceiling, the other an idle-gap detector.
  - Add `StreamMaxWaitExceededError` for the new failure mode (chunks still arriving under the inactivity gap, but total time exceeded the ceiling).
  - Drop the `PromptStream.replySubject` getter — under the mux inbox it loses its per-stream meaning. Documented as a debugging aid; no tests or examples in-tree consume
  it.

## Why

  The hand-rolled `subscribe → flush → publish` was a faithful re-implementation of what `requestMany` already does, plus extra round-trips (one SUB, one flush) per prompt
  and an extra inbox subject the broker has to track. The sentinel strategy lines up exactly with the protocol's terminator shape, so we get the same observable behavior
  with less code and less wire traffic.

## Notes for reviewers

  - `nc.requestMany` is typed in the `NatsConnection` interface as returning `AsyncIterable<Msg>`, but every concrete implementation returns the wider
  `QueuedIterator<Msg>`. We cast at the boundary with a comment, since `.stop()` is the only way to bail out early without waiting for `maxWait` to expire (used by
  `cancel()` and the abort-signal path).
  - `withInactivityTimeout` still wraps the iterator — `requestMany` only honors the absolute `maxWait`, not idle-gap reset.
  - Sentinel triggers on any empty-body message; the §9.1 error-signal (empty body + headers) ends iteration the same way as the §6.5 terminator. We still inspect the
  message inside the loop to decide whether to `throw ServiceError` or yield the synthetic `{status:"done"}`. The §9.3 trailing terminator after an error signal won't be
  observed, which is harmless — we're throwing anyway.
  - `examples/agent-web-ui/server/bridge.ts` references `StreamStalledError` only; it falls through to the generic `internal` handler for the new
  `StreamMaxWaitExceededError`. Out of scope here per the release ladder — the example pins `^0.4.x` and migrates after publish.

### DIFF
--- a/client-sdk/typescript/CHANGELOG.md
+++ b/client-sdk/typescript/CHANGELOG.md
@@ -13,6 +13,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+> **Release-tag note:** the `replySubject` removal under "Removed" below
+> is a breaking public API change (the getter was listed in the 0.4.0
+> public surface). Cut this as **0.5.0**, not a 0.4.x patch.
+
 ### Changed
 
 - `PromptStream` now drives the response with `nc.requestMany` using
--- a/client-sdk/typescript/CHANGELOG.md
+++ b/client-sdk/typescript/CHANGELOG.md
@@ -13,6 +13,33 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Changed
+
+- `PromptStream` now drives the response with `nc.requestMany` using
+  the `sentinel` strategy (the wire terminator's empty-body shape
+  doubles as the sentinel). Replies route through the connection's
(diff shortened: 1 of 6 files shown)


derekcollison:
@claude please check again - thanks.

derekcollison:
@claude please do one more review.

PR #67, opened 2026-05-03
Catch Python SDK up to TS PR #66 (mux reply-inbox + max_wait_s)

renerocksai:
## Summary

- **Catch the Python client SDK up to TypeScript [PR #66](https://github.com/synadia-ai/synadia-agents/pull/66)** (`requestMany` + sentinel) using a per-NATS-connection mux reply-inbox: every `Agent` on the same `nc` shares a single `_INBOX.agents.<mux>.*` subscription instead of opening one per prompt. **Wire shape unchanged; protocol stays at `"0.3"`**, the TS↔Python interop test continues to pass.
- **Add `Agent.prompt(max_wait_s=...)`** — absolute ceiling on a prompt stream, mirrors TS's `PromptOptions.maxWaitMs`. Default `600 s` (10 min). Distinct from the §6.6 per-chunk inactivity timer that resets on every received chunk; `max_wait_s` is the safety net for streams that emit a steady trickle forever, or for silent reconnect / broker-death windows.
- **New error classes**: `StreamMaxWaitExceededError` (ceiling), `StreamStalledError` (inactivity gap, previously a bare `ProtocolError`), `AgentsClosedError` (pre-flight close). All inherit from `ProtocolError` / `NatsAgentError` so existing catch-broadly callers keep working.

## Why interim

`nats-py` has no `request_many()` primitive yet; the TS SDK uses `nc.requestMany(...)` directly. We've filed [`nats-io/nats.py#919`](https://github.com/nats-io/nats.py/issues/919) requesting it. Until that lands, `synadia_ai.agents._mux` is the API-level analogue, intentionally mirroring TS's shape. Marker `INTERIM-NATSPY-REQUEST-MANY` in the source flags it for deletion when upstream catches up — the single migration point will be `Agent._stream_prompt`.

### Mux on the connection, not on `Agents`

The TS SDK's `nc.requestMany` lives on the connection — every caller of the same `nc` automatically shares the connection's internal mux. Python's analogue is a per-`nc` singleton `MuxInbox` held in a `WeakKeyDictionary` keyed by the connection (see `synadia_ai.agents._mux.mux_for`). Multiple `Agents` instances and directly-constructed `Agent` handles on the same connection share one subscription. `Agents.close()` does **not** tear down the mux — it lives on the connection and is the connection-owner's responsibility, just like in TS.

`MuxInbox` holds only a `weakref.ref(nc)` to the client (not a strong ref) so dropping the user's last `nc` reference frees the cache entry without manual teardown.

### Lifecycle events race queue reads

`close_event` and `max_wait_s` are lifecycle controls, not ordinary queued stream values. `_wait_for_chunk` does an `asyncio.wait` over `queue.get()` + `max_wait_event` + `close_event`, so close and max-wait win over already-buffered chunks (including a buffered terminator). Consumers unblock within an event-loop tick — no FIFO sentinel that could be hidden behind backlog.

The terminator detection runs in the mux's dispatch callback (sync-bodied, single-loop pass). When the empty-headerless `b""` terminator arrives on the wire, the mux cancels the per-stream `max_wait_handle` synchronously — not when a slow consumer eventually drains the terminator from the queue.

### Three close-event checks bracket every await before the publish

- Top of `_stream_prompt` (before `mux.start()`)
- Between `mux.register()` and `publish()` (catches close that fires during the first call's SUB+flush)
- After `publish()` (catches close fired during the publish itself; pushes sentinel synchronously)

All three raise `AgentsClosedError` from the pre-publish checks; the post-publish path joins the lifecycle race in `_wait_for_chunk`. Cleanup ordering in the `finally`: cancel close-watcher → unregister mux token → drain queue (releases `Msg` payloads deterministically rather than waiting on GC).

## Public API additions

```python
from synadia_ai.agents import (
    DEFAULT_PROMPT_MAX_WAIT_S,        # = 600.0
    StreamMaxWaitExceededError,        # ceiling fired (.max_wait_s)
    StreamStalledError,                # inactivity gap fired (.timeout_s, .reply_subject)
    AgentsClosedError,                 # called against closed Agents (or close fired during stream)
)

# Constructor default + per-call override:
agents = Agents(nc=nc, prompt_max_wait_s=300.0)
async for chunk in agent.prompt("...", max_wait_s=30.0):
    ...
```

`AgentsClosedError(NatsAgentError)` is **intentionally distinct** from `ProtocolError` so callers can branch on "called against a closed Agents" vs "torn down mid-flight" — see `errors.py:53`. Mid-stream close still surfaces as `ProtocolError`; pre-flight (and the now-narrowed mid-mux-start race window) raises `AgentsClosedError`.

## Test coverage

**203 → 207 tests, all green in 12.87 s.** New regression suites:

- `tests/test_prompt_max_wait.py` — 7 tests: ceiling fires on never-terminating stream, terminator-in-time completes cleanly, ceiling vs inactivity timer disambiguation (with elapsed-band + chunk-count assertions to pin the dual-timer claim), inactivity raises `StreamStalledError`, **broker-death mid-stream** (kills a dedicated nats-server, asserts ceiling rescues the stream), and a slow-consumer-with-buffered-terminator regression.
- `tests/test_prompt_cancel_race.py` — 7 tests: `Agents.close` mid-stream unblocks within ~50 ms (not on inactivity timer), 50 concurrent prompts on shared mux see only their own chunks, close-before-prompt raises `AgentsClosedError`, close-during-prompt raises `ProtocolError` (deterministic event-driven sync, not `sleep(0)`), close-wins-over-buffered-chunks-and-terminator, and **caller-side cancellation cleans up internal `agents-prompt-*` wait tasks** (asserts no orphan tasks via `asyncio.all_tasks()` + mux route is unregistered).
- `tests/test_mux_wire_economy.py` — 3 tests: SDK-instrumented (5 prompts → 1 SUB), **broker-observed via `/subsz` HTTP** (asserts the exact `<mux_prefix>.*` wildcard subject), and a weak-cache-cleanup regression (drops a `Client`, GC-collects, asserts the mux entry is gone).

The cross-SDK interop test (`tests/test_interop_e2e.py`) — Python client driving the TypeScript reference agent over wire — continues to pass.

## Test infrastructure

- `tests/harness/nats_server.py` now starts `nats-server` with `-m <port>` unconditionally and exposes `RunningServer.monitoring_url`. Lets tests assert broker-observed truth (e.g. subscription counts via `/subsz`) instead of monkey-patching SDK internals.
- `_pick_free_ports(n)` allocates `n` distinct free ports while holding all binds simultaneously — closes the TOCTOU window where two back-to-back `_pick_free_port()` calls could (vanishingly rarely) collide.
- `bg_tasks` fixture in `tests/conftest.py` cancels long-running emit-loops at teardown so `pytest-asyncio` doesn't print "Task was destroyed but it is pending."

## Lockstep version bumps

- `synadia-ai-agents` 0.5.0 → **0.6.0**
- `synadia-ai-agent-service` 0.1.0 → **0.2.0** (dep-pin only — the agent-side wire is unchanged)
- Both Python CHANGELOGs updated.

## Not changed

- Wire format. Protocol still `"0.3"`.
- TS↔Python interop. The TS SDK already shipped PR #66; the Python interop test (`tests/test_interop_e2e.py`) round-trips against the TS reference agent and stays green.

## Review history

This branch went through **two rounds of fresh-eyes review** (each round: independent implementation reviewer + independent test reviewer). Both reviewers landed on TRUST after their respective findings were addressed. Round-1 surfaced a queue-payload-retention bug, a `MuxInbox.start()` flush-failure leak, a third close-check gap, and a few tight-assertion tests; round-2 surfaced a `_pick_free_port` TOCTOU and the broker-observed wildcard-subject check. The senior reviewer (renerocksai, with the bot's hat off) then independently pushed a third round (`75c1551` and follow-ups) hardening the close-races-buffered-chunks contract, the terminator-on-arrival cancellation, the weak-ref-to-nc lifecycle, and a brief regression in `41fa33e` that was caught and fixed in `02dacf6`.

## Test plan

- [x] `uv run ruff check --no-cache .` — passes
- [x] `uv run ruff format --check .` — 42 files clean
- [x] `uv run mypy --no-incremental src tests examples` — no issues, 41 source files
- [x] `uv run pytest` — **207 passed in 12.87 s** (incl. TS interop)
- [x] Two rounds of fresh-eyes sub-agent review — both TRUST
- [ ] CI green
- [ ] Reviewer bot pass

### DIFF
--- a/agent-sdk/python/CHANGELOG.md
+++ b/agent-sdk/python/CHANGELOG.md
@@ -6,6 +6,27 @@ Format based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 This project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html);
 the 0.x line is explicitly unstable per protocol spec §11.2.
 
+## [0.2.0] - 2026-05-03
+
+### Changed
+
+- **Dependency floor bumped to `synadia-ai-agents>=0.6`** in lockstep
+  with the client-sdk's prompt-stream catch-up to the TS SDK's PR #66
+  (`requestMany` + sentinel) — see
+  [`client-sdk/python/CHANGELOG.md`](../../client-sdk/python/CHANGELOG.md)
+  `[0.6.0]` for the substance of that change. **No agent-side
+  code changes:** PR #66 was confirmed to touch only
+  `client-sdk/typescript/` (`gh pr view 66 --json files`); the
+  agent-host wire is identical pre/post. Agents still publish
+  individual chunks to `msg.reply` with the §6.5 zero-byte
+  terminator — whether the client subscribed per-stream or via a
+  shared mux is invisible from the agent's POV. The bump exists
+  purely to keep the published metapackage coherent so a user
+  installing `synadia-ai-agent-service` via PyPI pulls a client-sdk
+  that exposes the new `Agent.prompt(max_wait_s=...)` /
+  `StreamMaxWaitExceededError` / `StreamStalledError` surface shared
+  between both packages.
+
 ## [0.1.0] - 2026-04-30
 
 Initial release. **Carved out of `synadia-ai-agents` at the 0.5.0
--- a/agent-sdk/python/CLAUDE.md
+++ b/agent-sdk/python/CLAUDE.md
@@ -85,8 +85,9 @@ package).
   `parse_nats_url`, the `SERVICE_NAME` / `PROMPT_QUEUE_GROUP` constants,
   …) keeps its `from synadia_ai.agents import …` path unchanged.
 - **Dependency direction:** `synadia-ai-agent-service` depends on
-  `synadia-ai-agents` (`>=0.5` — the first release that ships the
-  trimmed surface). Not the other way around. The client-sdk must
+  `synadia-ai-agents` (`>=0.6` on the current release line; `0.5.0`
+  was the first release that shipped the trimmed surface). Not the
+  other way around. The client-sdk must
   remain installable and useful without ever pulling the agent-sdk.
 - **No code duplication.** Anything shared (envelope codec, subject
   sanitization, `HeartbeatPayload`, `_PROTOCOL_VERSION`, validation
--- a/agent-sdk/python/README.md
+++ b/agent-sdk/python/README.md
@@ -20,7 +20,7 @@ uv pip install -e .
 ```
 
 When both packages are on PyPI, plain `pip install
-synadia-ai-agent-service` will pull `synadia-ai-agents>=0.5`
+synadia-ai-agent-service` will pull `synadia-ai-agents>=0.6`
 automatically.
 
 ## Quickstart — host an agent
--- a/agent-sdk/python/pyproject.toml
+++ b/agent-sdk/python/pyproject.toml
@@ -37,7 +37,7 @@ Issues = "https://github.com/synadia-ai/synadia-agents/issues"
 # `uv sync`. This affects monorepo developers only — it is dev-time
 # config and is NOT persisted into the built wheel's METADATA, so an
 # external user installing `synadia-ai-agent-service` from PyPI gets
-# the declared dep (`synadia-ai-agents>=0.5`) from PyPI as normal.
+# the declared dep (`synadia-ai-agents>=0.6`) from PyPI as normal.
 synadia-ai-agents = { path = "../../client-sdk/python", editable = true }
 
 [dependency-groups]
--- a/agent-sdk/python/pyproject.toml
+++ b/agent-sdk/python/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "synadia-ai-agent-service"
-version = "0.1.0"
+version = "0.2.0"
 description = "Python agent-host SDK for the NATS Agent Protocol — register and serve spec-compliant agents over NATS"
 readme = "README.md"
 requires-python = ">=3.11"
@@ -23,7 +23,7 @@ classifiers = [
 dependencies = [
     "nats-py>=2.7",
     "pydantic>=2.5",
-    "synadia-ai-agents>=0.5",
+    "synadia-ai-agents>=0.6",
 ]
 
 [project.urls]
--- a/agent-sdk/python/uv.lock
+++ b/agent-sdk/python/uv.lock
@@ -383,7 +383,7 @@ wheels = [
 
 [[package]]
 name = "synadia-ai-agent-service"
-version = "0.1.0"
+version = "0.2.0"
 source = { editable = "." }
 dependencies = [
     { name = "nats-py" },
@@ -416,7 +416,7 @@ dev = [
 
 [[package]]
 name = "synadia-ai-agents"
-version = "0.5.0"
+version = "0.6.0"
 source = { editable = "../../client-sdk/python" }
 dependencies = [
     { name = "nats-py" },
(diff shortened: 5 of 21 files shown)


renerocksai:
Thanks for the review — went through all five findings, fixed three, rejected two with reasoning. Commit: 5c56673.

### 1. `agent.py:412` — `max_wait_s == 0` fires immediately ✅ fixed

You're right that `0` is a footgun. Rather than pick a "0 means unlimited" or "0 means immediate" semantic, I refused the value at the public boundary: `max_wait_s <= 0` now raises `ValueError` synchronously at every entry point — `Agent.prompt(max_wait_s=...)`, `Agent(prompt_max_wait_s=...)`, and `Agents(prompt_max_wait_s=...)`. There is no "no limit" sentinel because an unbounded prompt stream is the exact failure mode the ceiling exists to prevent. Pass `None` to use the default, or any strictly positive number.

Dropped the now-dead `else: max_wait_event.set()` branch and the `TimerHandle | None` annotation in `_stream_prompt` since the boundary check makes them unreachable. Added a parametrized test covering `0`, `0.0`, `-1`, `-0.5` against all three entry points.

### 2. `_mux.py:199` — multi-event-loop caveat ✅ fixed

Added a `.. note::` block to `mux_for`'s docstring documenting the contract: `_MUX_CACHE` is module-global and `MuxInbox` captures an `asyncio.Lock` at construction tied to whichever loop runs the first call. Sharing one `Client` across loops in different threads is not supported. Documentation contract, not a runtime check — the SDK is single-loop-asyncio in shape.

### 3. `agent.py:327` — per-chunk task creation ✅ documented as intentional

Looked at this carefully and concluded the lift-out optimization isn't worth it. Reasoning:

- The two event-wait tasks (`max_wait_event.wait()`, `_close_event.wait()`) target Events that fire **once each** and are terminal, so "re-create if fired" effectively means "create once at stream start." That's a real change, not just a code reshape.
- The current `finally` is the single place that guarantees no task outlives the read it served. Lifting the event-wait tasks into `_stream_prompt` breaks that locality and adds new ordering constraints between mux unregister, queue drain, and the lifted-task cancellations — all of which the close-race contract relies on.
- AI-stream chunk rates make the per-chunk `create_task` cost invisible next to model + wire latency. The `if not queue.empty()` fast path already creates zero tasks.

So I added a comment in `_wait_for_chunk`'s docstring documenting the intentionality, leaving a marker for whoever might revisit this if a high-frequency low-latency workload ever shows up on this SDK.

### 4. `_mux.py:146` — `_on_msg` should be sync `def` ❌ rejected

Verified against the installed `nats-py` source. The premise is wrong twice over:

- `nats/aio/subscription.py:210-213` (`Subscription._start`) explicitly raises `nats: must use coroutine for subscriptions` if `cb` is not `iscoroutinefunction`. Switching to `def` would crash, not run inline.
- `nats/aio/subscription.py:295-321` (`_wait_for_msgs`) dispatches via `await self._cb(msg)` inside a single long-lived per-subscription task. There is **no** per-message `asyncio.create_task` allocation — the proposed savings don't exist.

Code stays as `async def`.

### 5. `CHANGELOG.md:930` — compare URL uses `...HEAD` ❌ out of scope for this PR

`...HEAD` is correct Keep-a-Changelog convention until the `python-v0.6.0` tag exists. The fix belongs in the release commit (alongside the version-bump and `npm publish` / PyPI publish dance), not in this feature PR. Will land at release time.

---

Pre-push gate: `ruff check --no-cache`, `ruff format --check`, `mypy --no-incremental`, full pytest (211 passed including TS interop) all green.

renerocksai:
@claude, please review again — fixes pushed in 5c56673.

renerocksai:
Good catch on the residual dead guard — `max_wait_handle` is unconditionally a `TimerHandle` after the cleanup, so the `is not None` in the finally block was unreachable. Fixed in ef000c0.

renerocksai:
@claude, please review again — residual nit fixed in ef000c0.

PR #72, opened 2026-05-04
feat(agents/open-agent): inbound NATS bridge for vercel-labs/open-agents

renerocksai:
## Summary

New `@synadia-ai/open-agent` package that puts an open-agents `ToolLoopAgent` (read/write/edit/grep/glob/bash/task/skill/web_fetch/ask_user_question) on `agents.prompt.open-agent.<owner>.<session>`. Built on `AgentService` from `@synadia-ai/agent-service` — first agent in this repo to use it directly.

- `agents/open-agent/vendor/agent/` — verbatim copy of upstream `packages/agent/` at SHA `56ddf94`, plus the sandbox interface/types files. `tsconfig.json` `paths` rewrites `@open-agents/sandbox` to a custom barrel so the vendored agent code runs unmodified. Refresh procedure documented in `VENDORED.md`.
- `vendor/sandbox/local.ts` — `LocalSandbox` impl backed by `node:fs/promises` + `Bun.spawn` (50KB combined-output cap, `AbortSignal` composition for timeouts, env passthrough, `readFileBuffer`).
- `vendor/sandbox/factory.ts` — dispatches `state.type === "local"` to `connectLocalSandbox`; throws on `"vercel"` with a hint to use `examples/open-agent-vercel`.
- `src/bridge.ts` — `runBridge()` wires `AgentService` to a per-prompt `ToolLoopAgent`, lazy sandbox bootstrap, `splitResponseText` for long deltas, in-memory history, pluggable `agentFactory` for tests.
- `src/agent.ts` — replaces upstream client-side `ask_user_question` with one that round-trips through `PromptResponse.ask` (numeric/letter/substring matching; `decline` handling).
- `src/chunk-translator.ts` — UI-message-part → wire `Chunk[]` mapper with per-tool input/output summarisers.
- `src/model-factory.ts` — pluggable provider resolution. Ships `gatewayModelFactory()` (Vercel AI Gateway, default; carries upstream provider tuning) and `openRouterModelFactory()` (OpenRouter via OpenAI-compatible chat completions). CLI auto-selects based on which API key is set; `OPEN_AGENT_PROVIDER` forces a choice.
- `src/cli.ts` — flags + env, NATS context loading, SIGINT/SIGTERM graceful shutdown, stderr-only logging.

Companion `examples/open-agent-vercel/` runs the same `runBridge` with `@vercel/sandbox` swapped in — proves the sandbox seam is interchangeable. Vendors `packages/sandbox/vercel` since upstream isn't on npm yet.

## Out of scope (deferred)

Controller / spawn-per-session orchestrator, persistent history, sandbox isolation, status-code granularity beyond 500, attachments, the open-agents-side `nats_agent` outbound tool (Part C of the original design — separate PR).

## Test plan

- [x] `bun run typecheck` clean in `agents/open-agent` and `examples/open-agent-vercel`
- [x] `bun test` — 73 pass / 0 fail (31 own + 42 vendored upstream)
  - `test/local-sandbox.test.ts` — read/write/exec/stat/readdir/readFileBuffer
  - `test/chunk-translator.test.ts` — every UI-part type → expected chunks
  - `test/bridge.integration.test.ts` — embedded `nats-server` + stub agent factory; covers discovery, streaming + terminator, mid-stream `ask`, handler error → `ServiceError(500)`
- [x] Manual smoke: `bun run cli` against a real `nats-server` with OpenRouter (`OPENROUTER_API_KEY` + `anthropic/claude-sonnet-4`); round-tripped a file-creation prompt via `nats req`
- [ ] Manual smoke against `@vercel/sandbox` via `examples/open-agent-vercel/` — deferred (requires `VERCEL_TOKEN`; not blocking for this PR)


renerocksai:
@claude please re-review — pushed `5af0a2b` addressing all six findings from your first pass.

**Security (1, 2, 3 from your summary):**

1. **Path confinement in `LocalSandbox`.** Every FS method now resolves the input path and rejects anything outside `workingDirectory`; `exec` confines its `cwd` argument the same way. Vendored `read.ts` still resolves absolute paths through unmodified, but the fence at the sandbox boundary catches them regardless of which tool calls in. New tests cover `readFile` / `writeFile` / `readFileBuffer` against `/etc/passwd`-style paths, `..` traversal, and `exec` cwd. Symlink-crossing remains a documented gap; bash itself still runs with full host privileges, so the README "not isolated" disclaimer stands.
2. **Subprocess env allowlist.** `...process.env` is gone. Forwarded keys: `HOME`, `PATH`, `TERM`, `TMPDIR`, `USER`, `SHELL`, `LANG`, `LC_ALL`, `LC_CTYPE`, `PWD`. Caller-supplied `state.env` is merged on top. New test sets a fake `OPEN_AGENT_FAKE_API_KEY` in `process.env` and confirms `printenv` inside the subprocess doesn't see it.
3. **Subject-token validation.** `runBridge` rejects empty `owner` / `session` and anything matching `/[.*>\s]/` before registering the service. New tests cover `owner="evil.attacker"`, `session=">"`, and empty owner.

**Correctness (4):**

4. **`error` UIPart re-thrown.** The bridge throws as soon as it sees `part.type === "error"`, so `AgentService` maps it to §9.1 + the §6.5 terminator. New integration test injects an `error` part via the stub `agentFactory` and asserts the caller receives `ServiceError(500)` plus that text past the error never reaches the wire.

**Minor (5):**

5. **TOCTOU in `ensureSandbox`.** Switched to a memoised `Promise<SandboxBundle>` so two concurrent prompts collapse onto a single `sandboxFactory` call.

**Deferred (6):**

- Unbounded conversation history — consistent with v1 non-goals; truncation heuristic will land alongside persistence in a follow-up PR.

84 pass / 0 fail (was 73). Typecheck clean in both `agents/open-agent` and `examples/open-agent-vercel`.

renerocksai:
@claude another pass please — pushed two more commits on top of your last review.

**`4601a5c` — tool I/O on the wire as `status` chunks**

Tool calls now ride on §6.4 `status` chunks with a `<prefix>:<json>` payload, mirroring the convention `agents/claude-code` already uses (and that `examples/agent-web-ui/server/bridge.ts` already parses generically):

```text
status: "tool_use:{\"id\":\"<toolCallId>\",\"name\":\"bash\",\"input\":{\"command\":\"ls\"}}"
status: "tool_result:{\"tool_use_id\":\"<toolCallId>\",\"output\":\"exit 0\\nfoo.txt\",\"is_error\":false}"
```

Spec §6.4 explicitly requires callers to silently ignore unknown status values, so this is forward-compatible — `nats req` and other dumb subscribers see only the model's `response` text; rich UIs that opt into the convention get structured tool-call cards with no UI change. `tool_use_id` pairs `tool_use` → `tool_result` for the consumer. Tool-input-start is now silent (the announcement piggybacks on `tool_use`). Translator tests went from 12 → 21 cases.

**`04d012c` — bash `needsApproval` wired through the §7 query mechanism**

`bashTool({ needsApproval: bashApprovalViaNats(...) })` round-trips through `PromptResponse.ask`. Approval tokens: `y`/`yes`/`approve`/`approved`/`allow`/`ok`/`okay`/`1` (case-insensitive, whitespace-tolerant). Anything else → throws → AI SDK surfaces a `tool-output-error` the model can react to. Returning `true` from `needsApproval` would deadlock (no UI to grant approval), hence denial = throw. 14 unit tests in `test/agent.test.ts`.

**Intentional scope note — please don't flag these as findings:**

This bridge is a **demo** for `vercel-labs/open-agents`, packaging it as a NATS-protocol-callable agent **with its goodies and its flaws**. Two upstream behaviors are knowingly preserved:

1. **The dangerous-command heuristic is exactly the upstream list** (`/\brm\s+-rf\b/` and `lowerCommand.includes(".env")`, in `vendor/agent/tools/bash.ts:32-49`). It's fragile and easily bypassed (`rm -f *`, `mv * /tmp/x`, base64-decoded shell, etc.). We tested this — `rm -f *` is not flagged. We're not patching vendored code or reimplementing `bashTool` to broaden the list, because:
   - Our vendoring rule is "zero patches to upstream code."
   - Heuristic broadening is theater against an adversarial model. Real isolation is what the Vercel example provides.
   - The README at the top of the package already prominently warns that `LocalSandbox` is **not** an isolation boundary; bash runs as the bridge user with full host privileges. Use `examples/open-agent-vercel/` for real isolation.
2. The upstream `read.ts` absolute-path fallback still resolves absolute paths through. The path-confinement fence we added in `LocalSandbox` (commit `5af0a2b`) catches escapes regardless of which tool calls in, so the exposure is closed at the sandbox boundary even though vendored tool code is unchanged.

Both are documented behaviour, not regressions. Would still appreciate findings on:
- The new wire format (correctness, spec conformance, JSON-encoding edge cases)
- The bash approval flow (race conditions, AbortSignal handling on a denied query, timeout behaviour)
- Anything else that crosses the bridge ↔ vendored seam

108 pass / 0 fail, typecheck clean in both `agents/open-agent` and `examples/open-agent-vercel`.

renerocksai:
@claude one more verification pass — pushed `c191d8c` addressing Finding 1 and documenting Finding 3.

**Finding 1 — `tool_use` input size cap.** `summarizeToolInputForWire` in `chunk-translator.ts:163` elides `content` for `write` and `oldString`/`newString` for `edit` regardless of size, and falls through to a 1 KB-per-string default cap for any other tool. The wire status chunk for a 64 KB write now serializes well under 2 KB. 5 new translator tests cover the elision, the default cap, small-input pass-through, and the size guarantee.

**Finding 3 — AbortSignal gap.** Documented in the README's Permissions section as a known limitation: AI SDK's `tool()` hands `abortSignal` to `execute` but not to `needsApproval`, so an abandoned prompt during the approval wait hangs until `askUserQuestionTimeoutMs` fires. Lowering the timeout lowers the worst-case hang; a real fix needs upstream AI SDK changes.

**Finding 2 — explicitly deferred.** Building an integration test that exercises a thrown `needsApproval` end-to-end through a real `ToolLoopAgent` would require standing up a stub model + AI SDK harness for a single assertion about AI SDK contract behaviour. The unit tests cover our denial-via-throw path; the AI SDK behaviour is what we're trusting per its type signatures. Not adding the harness in this PR.

113 pass / 0 fail (was 108). Typecheck clean. Looking for confirmation that Finding 1's fix is tight (no missed input fields on other tools, the 1 KB cap is reasonable) plus anything new the diff introduced.

PR #76, opened 2026-05-04
Fix(python): restore `session` on $SRV.INFO metadata + §8.3/§8.7 payloads (#73)

### LINKED ISSUE #73: Python SDKs are missing the `session` field on `$SRV.INFO` metadata and §8.3 heartbeat / status payloads
## Summary

The Python `synadia-ai-agent-service` (host) and `synadia-ai-agents` (client) packages don't carry the `session` label in two places where the NATS Agent Protocol spec ([synadia-ai/nats-agent-sdk-docs](https://github.com/synadia-ai/nats-agent-sdk-docs)) calls for it. The TypeScript SDK already populates both, so this shows up as a Python-vs-TypeScript wire shape difference.

## Details

**1. `metadata.session` is not in `$SRV.INFO.agents` (§3.2).**

The host advertises `agent`, `owner`, and `protocol_version` in service metadata, but not `session`. §3.2 asks for the field on session-aware agents, with `"default"` as the value for session-less ones. Consumers that read `$SRV.INFO` to discover or filter agents by session don't see one when talking to a Python host.

**2. `session` is not in the §8.3 heartbeat or §8.7 status payloads.**

The `HeartbeatPayload` schema in the client SDK leaves the field out, with a comment noting that the heartbeat subject already contains the session token. The host's heartbeat publisher and §8.7 status responder also don't populate it. §8.3's payload table includes `session` on the wire when `metadata.session` is set.

This shows up in two places:

- Liveness trackers that want to key on `(agent, owner, session, instance_id)` need to fall back to parsing the subject. That's workable for heartbeats but less convenient for §8.7 status replies, which arrive on a request inbox rather than the heartbeat subject.
- The Python and TypeScript SDKs end up with different wire shapes for the same frame. The interop test in `client-sdk/python/tests/test_interop_e2e.py` doesn't currently assert on this field, which is why the difference wasn't caught.

## Why it's worth fixing

- The TypeScript SDK already populates both fields, and cross-language parity is one of the goals called out in the repo's CLAUDE.md.
- The change is backwards-compatible. Both SDKs use forward-compatible decoders (`extra="ignore"` on the Python side, the equivalent on the TS side), so older peers tolerate the new field, and newer peers tolerate its absence. `session_name` is already a required constructor argument on the Python host, so the value is available to emit without any caller-facing change.
- Session-less callers were already passing `session_name="default"`, which matches what §3.2 allows for that case.

---

Happy to open a PR for this if I can get write access to the repo, fork it, or pass the branch off to someone who already has it.

renerocksai:
Closes #73 — reported by @jarretlavallee.

> Note on history: an identical fix was pushed straight to `main` as
> commit `11fe3d2` earlier today by mistake (Claude Code session — my
> bad). That commit was reverted on `main` in `73ed51f` so this PR
> can carry the same content through normal review and pick up the
> reviewer bot like every other substantive change. The diff below
> is what would have been on `main` in the absence of the misstep,
> rebased on top of `2aeafa1` (the 0.5.0 release-prep wave that
> landed in between).

## Summary

- Restores `session` on `$SRV.INFO.metadata` (§3.2), the §8.3
  heartbeat payload, and the §8.7 status reply.
- The Python SDKs had dropped these during the 2026-04-28 session-
  name collapse, on the (wrong) reasoning that "the publishing subject
  IS the session." The spec disagrees: §3.2 + §8.3 + appendix
  B.11/B.11a all carry `session` on the wire, mirroring the 5th
  subject token (or `"default"` for session-less harnesses).
- The local CLAUDE.md / docs/protocol-mapping.md files codified the
  wrong shape, which is why neither I nor the reviewer bot caught it
  on the original collapse PR. Those are now fixed alongside the
  code so the next sweep doesn't re-validate the regression.

## What changed

Code:

- `agent-sdk/python` — `AgentService.start()` advertises
  `metadata.session = subject.session_name`; `build_heartbeat_payload`
  populates `session` (used by both the heartbeat publisher and the
  §8.7 status endpoint, so they emit the same shape).
- `client-sdk/python` — `HeartbeatPayload.session` restored as
  `str | None = None`. Optional on decode lets the SDK interop with
  spec-compliant session-less peers (e.g. a TS agent that omits
  `options.session`); the agent-host always populates it on emit.

Tests:

- Flipped the negative regression guards (`assert "session" not in
  metadata` / `not in encoded`) into positive assertions on the
  field.
- Added `test_decoder_tolerates_missing_session` to lock in the
  optional-decode contract.
- Round-tripped the canonical §8.3 / appendix B.11 wire example.

Versioning:

- `synadia-ai-agents` 0.6.0 → 0.7.0 (new optional field on the
  shared wire model).
- `synadia-ai-agent-service` 0.2.0 → 0.3.0 with floor bumped to
  `synadia-ai-agents>=0.7` so the publisher can populate the field.

Docs:

- Both Python `CLAUDE.md` files corrected (no more "exactly three
  fields", no more "no `session` on the payload").
- `client-sdk/python/docs/protocol-mapping.md` rows for §3.2 and
  §8.3 updated.
- CHANGELOG entries on both packages.

## Test plan

- [x] `uv run pytest` in `client-sdk/python` — 211 passed (incl.
      `test_interop_e2e.py` against the TS reference agent).
- [x] `uv run pytest` in `agent-sdk/python` — 38 passed.
- [x] `uv run ruff check --no-cache .` clean in both packages.
- [x] `uv run ruff format --check .` clean in both packages.
- [x] `uv run mypy --no-incremental src tests examples` clean in both.
- [x] Inspected `tests/_evidence/.../srv-info.json` — confirms
      `metadata.session` lands on the wire.
- [x] Inspected `tests/_evidence/.../heartbeat.json` — confirms
      `session` lands on the heartbeat frame.

### DIFF
--- a/agent-sdk/python/CHANGELOG.md
+++ b/agent-sdk/python/CHANGELOG.md
@@ -6,6 +6,37 @@ Format based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 This project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html);
 the 0.x line is explicitly unstable per protocol spec §11.2.
 
+## [0.3.0] - 2026-05-04
+
+Restores wire-shape parity with the spec and TS SDK after the
+2026-04-28 session-name collapse mistakenly dropped `session` from
+service metadata and §8.3 / §8.7 payloads. Reported as
+[issue #73](https://github.com/synadia-ai/synadia-agents/issues/73).
+
+### Fixed
+
+- **`metadata.session` (§3.2)** — `AgentService.start()` now advertises
+  `metadata.session = session_name` alongside the existing
+  `{agent, owner, protocol_version}` triple. Per §3.2 a session-less
+  harness MAY omit the field OR set it to `"default"`; since the
+  Python constructor takes a required `session_name` (callers pass
+  `"default"` when session-less), we always emit it for a uniform
+  shape across both styles.
+- **`HeartbeatPayload.session` (§8.3)** — `build_heartbeat_payload`
+  now populates `session=subject.session_name`, so periodic
+  heartbeats published on `agents.hb.{a}.{o}.{session_name}` mirror
+  `metadata.session` per §8.3 / appendix B.11.
+- **Status reply `session` (§8.7)** — the same builder feeds the
+  `agents.status.{a}.{o}.{session_name}` request/reply endpoint, so
+  a §8.7 status reply carries `session` matching the heartbeat. §8.7
+  + appendix B.11a explicitly require the same §8.3 schema.
+
+### Changed
+
+- **Dependency floor bumped to `synadia-ai-agents>=0.7`** so the
+  shared `HeartbeatPayload` model has the `session` field — needed
+  for the publisher and status handler to populate it.
+
 ## [0.2.0] - 2026-05-03
 
 ### Changed
--- a/agent-sdk/python/CLAUDE.md
+++ b/agent-sdk/python/CLAUDE.md
@@ -158,11 +158,14 @@ deferred to a follow-up. All shapes are imported from
   wants to host N sessions registers N services. The agent-sdk picks
   the session names; the client filters on them.
 - **Service registration** (§3): every agent registers as a NATS micro
-  service named `agents` with `metadata = {agent, owner,
-  protocol_version = "0.3"}` — exactly three fields under v0.3. The
-  `prompt` endpoint MUST be registered with queue group `"agents"`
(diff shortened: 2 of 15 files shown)


renerocksai:
Claude code just went completely bonkers. It pushed to main again instead of this PR branch

renerocksai:
@claude so sorry, but my local Claude Code suddenly got extremely dumb. It pushed to main directly instead of staying in this PR. What I would like you to do is to review the following commit [9bd6e4b26c0b55b04860af2910210669908edfd1 as if it were part of this PR and flag anything that still needs fixing.

PR #86, opened 2026-05-04
agent-web-ui: visual restructure + multi-select fan-out (Phase 1)

M64GitHub:
## Summary

Five small steps that together move the dashboard from a per-bucket list view into a cleaner two-section grid with multi-prompt fan-out:

- **Top-row CONTROLLER badge** on pi-headless / cc-headless cards — violet outlined pill complementing the existing tag.
- **Two-section grid**: "Agents / Sessions" (everything promptable, alphabetical by family with agent before session-of-same-family) and "Controllers" (pi-headless + cc-headless only). Empty sections hide entirely; previous per-bucket list collapsed.
- **Equalised card heights per row** + a `flex: 1` spacer above the subject, so the head/title/meta/cwd block anchors to the top while subject/stats/badges/hint dock to the bottom.
- **`startPromptStream` helper** extracted out of `ChatPanel.onSubmit` (no behaviour change for ChatPanel; lets the new bar share the chat-mirroring logic without ~80 lines of duplication).
- **Phase 1 of multi-select fan-out**: a `Set`-based selection store, a hover-revealed circle on each promptable card (controllers + expired sessions excluded), and a slide-up action bar with prompt textarea + attach + Send-to-N + clear-×. Send fans the same prompt out to every selected non-busy agent; each stream lands in its own per-instance chat. Inline report after send: "Sent to 4 of 5 (1 busy)". Esc clears. Selection is in-memory and orthogonal to the right-panel chat.

Phase 2 (the "virtual session" that aggregates streams from N agents into one transcript) is queued for the next PR — the disabled toggle in the bar is its placeholder so the layout doesn't shift when it lands.

## Test plan

- [ ] Top-row badge: only pi-headless and cc-headless **controllers** show CONTROLLER. Sessions and other agent kinds don't.
- [ ] Sections: with mixed agents online, the order in "Agents / Sessions" is Claude Code → CC headless sessions → Hermes → Open Agent → OpenClaw → PI → PI headless sessions. Empty sections vanish.
- [ ] Card heights: in a row mixing a plain Hermes/Claude-Code agent and a CC-headless session with cost + lifetime, all cards in that row are the same height. Subject + badges sit at the bottom edge of each card.
- [ ] Multi-select: hover reveals the circle on agents/sessions, never on controllers and never on expired sessions. Click a few, the bar slides up; Esc dismisses; the × dismisses; clearing while sending works without crashing.
- [ ] Fan-out: with 3+ targets selected, "Send to N" produces a streaming response in each target's own chat panel. With one target mid-prompt, the busy count is correct and the inline report says "Sent to N-1 of N (1 busy)".
- [ ] Right-panel chat keeps working independently while multi-select is active.
- [ ] `claude / Automatic` reviewer bot has nothing load-bearing to flag.

### DIFF
--- a/examples/agent-web-ui/src/components/AgentCard.vue
+++ b/examples/agent-web-ui/src/components/AgentCard.vue
@@ -11,6 +11,7 @@ import {
 } from "../stores/agents.ts";
 import { onStopped, piexecState } from "../stores/piexec.ts";
 import { ccexecState, onCcStopped } from "../stores/ccexec.ts";
+import { selectionState, toggleSelection } from "../stores/selection.ts";
 import { useBridge } from "../composables/useBridge.ts";
 
 const props = defineProps<{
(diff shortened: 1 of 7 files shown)


M64GitHub:
@claude - great feedback. we resolved, please have a second look.

PR #87, opened 2026-05-04
agent-web-ui: virtual sessions — aggregate fan-out chats (Phase 2)

M64GitHub:
## Summary

Builds on PR #86's multi-select fan-out. Ticking the new "Stream all to a virtual session" toggle in the multi-select bar + Send creates a persistent UI-only **virtual session** locked to the current selection: every prompt fans out to the same N agents and aggregates the streamed responses into one transcript. State is in-memory only — refreshing the page clears virtual sessions.

- **New top section "Virtual Sessions"** above Agents / Sessions in the grid; hidden when none exist (matches our hide-empty-section rule). Fuchsia palette entry (`--bucket-virtual`) so virtual cards are distinct from agent / controller / session families.
- **Per-card UI**: VIRTUAL pill, label ("Virtual #1 (3 agents)"), online-target count, full target list (offline targets are line-through), and a trash button that cancels in-flight streams before deletion.
- **VirtualChatPanel** opens in the right panel when a virtual card is selected. One linear transcript with: one user bubble per turn, then one streaming bubble per non-busy target. Per-source headers ("@derek · CLAUDE-CODE") render between bubbles when the source changes — they don't repeat across consecutive bubbles from the same source (so a tool-use → tool-result → agent-text sequence from one agent stays under a single header).
- **Streams mirror to both** the per-instance chat and the virtual transcript, so an agent's own ChatPanel stays consistent if you click into it. Tool calls, query bubbles, per-turn cost, attachments, and stop are all wired up.
- **Offline / busy targets** at send time get a `(offline — skipped)` / `(busy — skipped)` placeholder bubble in the virtual transcript and no real prompt fires. Subsequent prompts (after the busy ones free up) fan out normally.

## Implementation notes

- `stores/virtualSessions.ts`: reactive `Map` of sessions keyed by `virtual:<uuid>`. The id prefix lets the existing `agentsState.selectedInstanceId` address either a real agent or a virtual session; `agents.ts setAgents` skips clearing the selection when the id is a virtual one.
- `composables/virtualPromptStreaming.ts`: `startVirtualTurn` classifies each target (online/offline/busy) and `fanoutOneTarget` fires the actual `bridge.prompt`, with every callback mirroring to both stores. Same `syncErrored` guard as the single-agent helper.
- `MultiSelectBar`: the toggle is enabled, sends now branch on `virtualMode`. Send button gains a fuchsia gradient when virtual mode is on; placeholder/label text adapts.
- Right panel selection routing is `selectedVirtualSession ?? selectedAgent ?? empty-state` — virtual takes precedence because virtual ids never resolve to a real agent.

## Test plan

- [ ] Tick at least 2 selectable cards → tick "Stream all to a virtual session" → type a prompt → Send. A new "Virtual Sessions" section appears at the top of the grid; the right panel opens the new virtual chat with one user bubble + one per-source streaming bubble per target.
- [ ] Tool-using agent (e.g. claude-code-headless): tool-use + tool-result bubbles render under the same source header without repeating it.
- [ ] Type a follow-up prompt inside the virtual chat → fans out to the same locked targets. Source headers reset per turn.
- [ ] One target busy at send time: a `(busy — skipped)` placeholder appears in the virtual transcript; the real prompt is not fired (per-instance chat is untouched).
- [ ] One target removed (controller stops the session, or agent disconnects): subsequent virtual turn skips it with `(offline — skipped)`.
- [ ] Click into one of the per-instance chats while a virtual turn is streaming: same response is mirrored there.
- [ ] Stop button in the virtual chat cancels every per-source stream of the active turn.
- [ ] Trash button on a virtual card cancels in-flight streams + removes the session + clears the right-panel selection if it was viewing that virtual session.
- [ ] Plain (non-virtual) fan-out from Phase 1 still works exactly as before when the toggle is off.

### DIFF
--- a/examples/agent-web-ui/src/components/AgentGrid.vue
+++ b/examples/agent-web-ui/src/components/AgentGrid.vue
@@ -2,11 +2,18 @@
 import { computed } from "vue";
 import AgentGroup from "./AgentGroup.vue";
 import MultiSelectBar from "./MultiSelectBar.vue";
+import VirtualSessionsSection from "./VirtualSessionsSection.vue";
 import { agentsState, agentSections } from "../stores/agents.ts";
 import { selectionState } from "../stores/selection.ts";
+import { virtualSessionsList } from "../stores/virtualSessions.ts";
 
 const groups = computed(() => agentSections.value);
-const isEmpty = computed(() => groups.value.length === 0);
+// Empty state for the placeholder card only fires when there are no real
+// agents AND no virtual sessions; a fresh page with virtual sessions
+// alone still needs the grid header chrome.
+const isEmpty = computed(
+  () => groups.value.length === 0 && virtualSessionsList.value.length === 0,
+);
 const hasSelection = computed(() => selectionState.ids.size > 0);
 </script>
 
@@ -33,6 +40,7 @@ const hasSelection = computed(() => selectionState.ids.size > 0);
         </p>
       </div>
       <div v-else class="groups">
+        <VirtualSessionsSection />
         <AgentGroup
           v-for="g in groups"
           :key="g.id"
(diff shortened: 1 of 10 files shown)


M64GitHub:
@claude - please review again, we fixed the issues.

PR #92, opened 2026-05-05
agents/openclaw: auto-bootstrap channels.nats from setRuntime (0.5.4)

M64GitHub:
## Summary
- After fresh `openclaw plugins install @synadia-ai/nats-channel`, the plugin loaded into the runtime registry but the gateway never started a NATS channel — `channels.nats` was missing from `openclaw.json`, so the gateway's channel iteration skipped it. End-user symptom: install succeeded, env vars set, gateway started, **no NATS connection**.
- **Root cause**: openclaw 2026.5.4 fires `registerFull` only for plugins it considers "configured" (i.e., `channels.<id>` already present in cfg). For npm-installed channel plugins not in openclaw's official catalog, the loader uses **"discovery" mode** — `registerChannel` + `setRuntime` fire, `registerFull` doesn't. Our `ensureNatsChannelConfig` was wired only to `registerFull`, so the bootstrap step never ran.
- **Fix**: also call `ensureNatsChannelConfig` from `setRuntime`, which fires in discovery mode (proof: `openclaw plugins inspect nats` reports `Capabilities: channel: nats` — that means `registerChannel` ran, and per the SDK, `setRuntime` always runs alongside `registerChannel`). `registerFull` keeps the call as belt-and-suspenders for the full-mode path that bundled / catalog-known plugins use. The function is idempotent and fast, so re-running across modes is safe.
- **User-visible effect**: post-merge + publish, the entire quickstart becomes:
  ```sh
  openclaw plugins install @synadia-ai/nats-channel
  export NATS_CONTEXT=ngs NATS_AGENT_NAME=my-agent NATS_OWNER=my-org
  openclaw gateway
  ```
  No manual `openclaw.json` edit. No wizard. No symlink workaround.

## Implementation notes
- `ensureNatsChannelConfig(api)` → `ensureNatsChannelConfig(runtime)` since `setRuntime` only receives the `PluginRuntime`. Logging falls back to `console.*` because `api.logger` isn't available in the `setRuntime` callback.
- Wraps the previously-unawaited `writeConfigFile` Promise in a `.catch` so persistence errors surface instead of silently triggering an unhandled rejection.

## What this does *not* fix
- The wizard (`openclaw configure --section channels`) still won't show NATS — that requires either openclaw catalog enrollment of `@synadia-ai/nats-channel` or an upstream fix to `listChannelCatalogEntries` (it doesn't pass `installRecords` to `discoverOpenClawPlugins`). Both are separate openclaw PRs and tracked elsewhere.

## Test plan
- [ ] After publish to npm, fresh install in a clean openclaw container:
  - [ ] `openclaw plugins install @synadia-ai/nats-channel` succeeds
  - [ ] `[nats] created default channel config entry` appears in startup logs
  - [ ] `channels.nats.accounts.default = {}` lands in `~/.openclaw/openclaw.json`
  - [ ] With `NATS_CONTEXT=ngs NATS_AGENT_NAME=demo NATS_OWNER=m64`, `openclaw gateway` brings up the channel and `nats micro list` (or `$SRV.INFO.agents`) sees the agent
  - [ ] Re-running `openclaw plugins install` or `openclaw gateway` is idempotent (no duplicate writes, no errors)

## Caveat — runtime not yet verified
This is a code-trace fix: the openclaw SDK clearly invokes `setRuntime` whenever `registerChannel` fires, and `inspect nats` confirms the latter for our plugin. But we have not actually restarted a gateway with the new dist and observed the bootstrap log line. End-to-end validation will happen post-publish from npm.

### DIFF
--- a/agents/openclaw/index.ts
+++ b/agents/openclaw/index.ts
@@ -1,4 +1,4 @@
-import type { ChannelPlugin, OpenClawPluginApi } from "openclaw/plugin-sdk/core";
+import type { ChannelPlugin, OpenClawPluginApi, PluginRuntime } from "openclaw/plugin-sdk/core";
 import { defineChannelPluginEntry } from "openclaw/plugin-sdk/core";
 import { natsPlugin } from "./src/channel.js";
 import { setNatsRuntime } from "./src/runtime.js";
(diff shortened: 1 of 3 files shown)


M64GitHub:
@claude - we addressed the issues - please review again

PR #99, opened 2026-05-11
Docs: nats-req --reply-timeout gotcha for streaming agents

M64GitHub:
## Summary

`nats req --wait-for-empty` (and `--replies=0`) has a hidden default of `--reply-timeout=300ms` — the max gap allowed between consecutive replies. `pi`, `openclaw`, and `pi-headless` publish an immediate `{type:"status",data:"ack"}` chunk on request receipt, but the LLM's first response chunk typically lands 1–2 s later. The default fires in the ack→first-response gap and the CLI exits with just the ack, making the agent look broken from the CLI even though SDK callers see it stream cleanly.

This PR updates the three affected READMEs to pass `--reply-timeout 30s` in their `nats req` examples and adds a short note explaining the gotcha.

## Root cause walk-through

| Agent | First reply | Gap to first response | `nats req --wait-for-empty` outcome |
| --- | --- | --- | --- |
| **pi** | `{type:"status",data:"ack"}` at t≈100 ms | ~1.5 s | CLI exits after ack (default 300 ms) |
| **openclaw** | `{type:"status",data:"ack"}` at t≈100 ms | varies (LLM warmup) | CLI exits after ack |
| **hermes** (v0.2 fork) | first response chunk directly | n/a | streams cleanly |
| **claude-code** | first response chunk directly (ack is keep-alive only, 30 s interval) | n/a | streams cleanly |
| **open-agent** | same as claude-code | n/a | streams cleanly |

SDK callers (`nc.requestMany` with `strategy:"sentinel"`) wait the full `maxWait` regardless of inter-arrival gaps, which is why the web UI and all the example scripts work fine — only the Go CLI defaults bite.

## What changed

- `agents/pi/README.md` — updated the "Talk to your session" examples and added a troubleshooting entry
- `agents/openclaw/README.md` — same
- `examples/pi-headless/README.md` — updated the Prompt example

## Not affected

- `agents/claude-code/README.md`, `agents/open-agent/README.md` — these agents have no immediate ack, so the CLI examples work as-is.
- Root `README.md` — `nats req` is only used there for discovery (`$SRV.INFO.agents` with `--replies=0 --timeout=2s`), which is single-shot.

## Test plan

- [x] Verified the fix end-to-end against the user's deployed `pi` and `openclaw` on NGS — `nats req --wait-for-empty --reply-timeout 30s --timeout 60s` now streams the full response and terminates on the empty body.
- [x] Confirmed the issue does NOT exist for `claude-code`, `open-agent`, and `hermes` (different ack patterns).
- [x] No code changes — docs only.

### DIFF
--- a/agents/openclaw/README.md
+++ b/agents/openclaw/README.md
@@ -188,19 +188,23 @@ From the CLI:
 
 ```bash
 # Plain text prompt
-nats req agents.prompt.oc.<owner>.<agentName> "Hello!" --wait-for-empty --timeout 60s
+nats req agents.prompt.oc.<owner>.<agentName> "Hello!" \
+  --wait-for-empty --reply-timeout 30s --timeout 60s
 
 # JSON envelope (caller SDKs use this form under the hood)
-nats req agents.prompt.oc.<owner>.<agentName> '{"prompt":"Hello!"}' --wait-for-empty --timeout 60s
+nats req agents.prompt.oc.<owner>.<agentName> '{"prompt":"Hello!"}' \
+  --wait-for-empty --reply-timeout 30s --timeout 60s
 
 # With an attachment
 nats req agents.prompt.oc.<owner>.<agentName> '{
   "prompt": "describe this image",
   "attachments": [{"filename":"pic.png","content":"<base64>"}]
-}' --wait-for-empty --timeout 120s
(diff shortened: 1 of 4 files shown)


M64GitHub:
@claude - please final review (I hope) !

PR #100, opened 2026-05-11
AgentService (TS): emit §6.4 mandatory leading ack before handler runs

M64GitHub:
## Summary

TS-side counterpart to [PR #101](https://github.com/synadia-ai/synadia-agents/pull/101) (Python SDK, merged) — brings `@synadia-ai/agent-service` into parity with the §6.4 spec tightening ([nats-agent-sdk-docs@b1c6972](https://github.com/synadia-ai/nats-agent-sdk-docs/commit/b1c6972)), which requires an immediate `{"type":"status","data":"ack"}` chunk as the first message on the reply subject, before any latency-inducing handler work.

> **Rebased and rescoped.** This branch was rebased onto `main` after PRs #101 + #102 landed; the Python changes that lived in earlier revisions of this branch are now in main and have been dropped. What remains is the TS-only change.

## What changes

- `agent-sdk/typescript/src/service.ts` — `#dispatchPrompt` now publishes `encodeChunk({type:"status",status:"ack"})` synchronously after envelope decode and before invoking the handler. Comment block uses "leading ack" terminology for cross-SDK consistency with the Python side. The pre-existing periodic keep-alive interval is retained as §6.6 inactivity-timer defense; `keepaliveIntervalS` now controls only the periodic cadence, not the leading ack.
- `agent-sdk/typescript/test/integration/agent-service.test.ts` — new test `emits the §6.4 mandatory leading 'ack' as the first chunk before the handler runs`: handler blocks on a promise, the test asserts the ack lands before the handler is released. `try/finally` around the pre-release assertions guarantees the suspended handler is unblocked even on assertion failure.
- `agent-sdk/typescript/CHANGELOG.md` — `[Unreleased]` entry mirroring the Python SDK's `agent-sdk/python/CHANGELOG.md` shape.

## Wire compatibility

Fully compatible. Callers already decode arbitrary `status` chunks via `@synadia-ai/agents`. The only observable change is one extra `ack` event at the head of every stream, which clients have always accepted.

## Test plan

- [x] `agent-sdk/typescript` typecheck + eslint + prettier `--check` + 38 tests (37 existing + 1 new) all green against a local NATS broker
- [x] `client-sdk/typescript` 280 tests still pass — wire decode untouched
- [x] Verified the new test fails fast (assertion-error, not timeout) if the eager ack is removed from `#dispatchPrompt`

### DIFF
--- a/agent-sdk/typescript/CHANGELOG.md
+++ b/agent-sdk/typescript/CHANGELOG.md
@@ -6,6 +6,28 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Changed
+
+- **Leading `status=ack` chunk is now emitted unconditionally (§6.4).**
+  Spec §6.4 was sharpened to require that every prompt handler emit
+  exactly one `{"type":"status","data":"ack"}` chunk as the **first**
+  message on the reply subject, **before** any work that introduces
+  observable latency
+  ([nats-agent-sdk-docs@b1c6972](https://github.com/synadia-ai/nats-agent-sdk-docs/commit/b1c6972)).
+  `AgentService.#dispatchPrompt` now publishes the ack after a
+  successful envelope decode and before invoking the user-supplied
+  handler — so every TS agent built on `AgentService` (the reference
+  agent, `agents/open-agent`, any third-party harness) becomes
+  spec-compliant on upgrade with no code change. The ack is emitted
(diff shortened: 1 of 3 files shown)


M64GitHub:
@claude - please review again, I think we fixed everything

M64GitHub:
@claude - we had to abandon the python stuff, as that is already done meanwhile by PR 101 + PR 102. Can you please check now.

PR #101, opened 2026-05-11
Python SDK: §6.4 mandatory leading ack

renerocksai:
## Summary

- Spec §6.4 was clarified: agents MUST emit exactly one `{"type":"status","data":"ack"}` chunk as the **first** message on the reply subject, before any `response`/`query` chunk and before any work that introduces observable latency. The ack resets the caller's §6.6 inactivity timeout ahead of any warm-up gap and makes the stream observable to generic NATS tooling (`nats req --wait-for-empty`).
- Implements the §6.4 invariant at a single chokepoint: `AgentService._on_prompt_request` now publishes the leading ack unconditionally after a successful envelope decode and before invoking the user handler. Every Python agent in the repo (reference agent, `demo_echo`, in-tree test handlers) becomes spec-compliant on upgrade with no per-agent code change.
- Leading ack is independent of `keepalive_interval_s` — passing `None` only disables the periodic keep-alive cadence. Malformed envelopes still produce `error(400) → terminator` with no spurious ack (ack lives after decode validation).

## Changes

- **agent-sdk**: 5-line edit to `service.py` inserting the ack between decode and handler; one comment-block tweak documenting the relationship to keep-alive.
- **agent-sdk tests**: new `test_leading_ack_e2e.py` covers the wire-shape ordering and the no-ack-before-400 invariant. Existing tests rewritten for the new invariant — see commit-by-commit history.
- **client-sdk**: no functional change. `06-chat.py` spinner trigger now waits for the first `ResponseChunk`/`Query` rather than the first chunk overall, so the "thinking…" spinner keeps spinning through the leading ack instead of stopping prematurely.
- **docs**: `CHANGELOG.md` entries on both packages + a 2026-05-11 alignment milestone in `client-sdk/python/CLAUDE.md`.

## Pairing with TS

This is the Python half. The TS half is being landed in parallel by Mario (branch `sdk-mandatory-initial-ack`). The client side of either SDK already tolerates ack absence, so the cross-SDK interop test stays green throughout the rollout — there's a brief asymmetry on the wire while only one SDK has shipped, but no regression.

## Test plan

- [x] `cd agent-sdk/python && uv run ruff check --no-cache .`
- [x] `cd agent-sdk/python && uv run ruff format --check .`
- [x] `cd agent-sdk/python && uv run mypy --no-incremental src tests examples`
- [x] `cd agent-sdk/python && uv run pytest -v` — 40 passed
- [x] `cd client-sdk/python && uv run ruff check --no-cache .`
- [x] `cd client-sdk/python && uv run mypy --no-incremental src tests examples`
- [x] `cd client-sdk/python && uv run pytest -v` — 212 passed (including `tests/test_interop_e2e.py` against the unchanged TS reference agent — confirms the client receive side tolerates ack absence in the asymmetric window before the TS fix lands)
- [x] Wire trace evidence: `agent-sdk/python/tests/_evidence/tests__test_leading_ack_e2e.py__test_prompt_emits_leading_ack/frames.jsonl` shows `status=ack` first, then the handler's response, then the §6.5 terminator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Both languages stay in lockstep on the wire format, validated by a cross-SDK int
        └─── streamed chunks ───────┘
 ```
 
-A request is plain UTF-8 text or a JSON envelope `{"prompt": "...", "attachments": [{"filename": "...", "content": "<base64>"}]}`. The agent streams typed JSON chunks on the reply subject — `{"type":"response","data":"..."}` for content, `{"type":"status","data":"ack"}` for keep-alive, `{"type":"query","data":{...}}` for mid-stream questions — and ends with an **empty-body, no-headers** terminator. Errors use the `Nats-Service-Error-Code` header (`400` client, `500` server).
+A request is plain UTF-8 text or a JSON envelope `{"prompt": "...", "attachments": [{"filename": "...", "content": "<base64>"}]}`. The agent streams typed JSON chunks on the reply subject — `{"type":"response","data":"..."}` for content, `{"type":"status","data":"ack"}` as the mandatory §6.4 leading chunk (and optionally again as periodic keep-alive), `{"type":"query","data":{...}}` for mid-stream questions — and ends with an **empty-body, no-headers** terminator. Errors use the `Nats-Service-Error-Code` header (`400` client, `500` server).
 
 Discovery is standard NATS micro:
 
--- a/agent-sdk/python/CHANGELOG.md
+++ b/agent-sdk/python/CHANGELOG.md
@@ -6,6 +6,24 @@ Format based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 This project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html);
 the 0.x line is explicitly unstable per protocol spec §11.2.
 
+## [Unreleased]
+
+### Changed
+
+- **Leading `status=ack` chunk is now emitted unconditionally (§6.4).**
+  Spec §6.4 was sharpened to require that every prompt handler emit
+  exactly one `{"type":"status","data":"ack"}` chunk as the **first**
+  message on the reply subject, **before** any work that introduces
+  observable latency. `AgentService._on_prompt_request` now publishes
+  the ack after a successful envelope decode and before invoking the
+  user-supplied handler — so every Python agent in the repo
+  (reference agent, `demo_echo`, in-tree test handlers) becomes
+  spec-compliant on upgrade with no code change. The ack is emitted
+  regardless of `keepalive_interval_s`; passing `None` only disables
+  the periodic keep-alive cadence, not the leading ack. A malformed
+  envelope still produces `error(400) → terminator` with no
+  spurious ack — the ack lives after decode validation.
+
 ## [0.3.0] - 2026-05-04
 
 Restores wire-shape parity with the spec and TS SDK after the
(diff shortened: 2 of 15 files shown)


renerocksai:
@claude, please review again — addressed both of your findings plus three more from a post-PR audit.

**Your Finding 1** (consolidate `_collect_replies`): extended the helper to take an optional `payload`, refactored `test_malformed_envelope_emits_400_then_terminator` onto it. Both error-completion tests now use the same collection logic and stop on the first §6.5 terminator (no `len >= 2` guard).

**Your Finding 2** (`AgentService` docstring): updated the class docstring for `keepalive_interval_s=None` to make the leading-vs-periodic distinction explicit — agent authors reading `help(AgentService)` or an IDE hover now get the full story, not just the periodic-cadence-only story.

**Three additional fixes from the audit:**

- `client-sdk/python/tests/test_interop_e2e.py::test_python_client_prompts_ts_reference_agent` asserted `isinstance(msg, ResponseChunk)` on every chunk in the TS reference agent's stream. Passes today (TS reference agent hasn't been updated yet) but would fail the moment the parallel `sdk-mandatory-initial-ack` branch lands. Accept `StatusChunk` in the iterator and filter ResponseChunks for the content assertion.
- `test_keepalive_emits_only_leading_ack_for_fast_handler` counted acks but didn't assert the ack was the FIRST chunk under keep-alive-enabled. Pinned `received[0]` to mirror the keep-alive-disabled sibling so a regression that emits the ack AFTER a response chunk can't slip through the count assertion.
- Doc drift in `client-sdk/python/docs/protocol-mapping.md` (parenthetical "§6.4 status chunks are at the agent's discretion" — stale for the leading ack) and `README.md` (characterised ack as "for keep-alive" only). Both reworded to make the leading-vs-periodic split explicit.

Two commits on top of the original five:
- `35918de tests: tighten coverage + fix interop test for incoming TS leading ack`
- `0620a23 docs: clarify leading vs periodic ack across user-facing surfaces`

Pre-push gate green on both SDKs: ruff (no-cache) ✓ · ruff format ✓ · mypy (no-incremental) ✓ · pytest 40 ✓ (agent-sdk) · pytest 212 ✓ (client-sdk, incl. interop).

renerocksai:
@claude, please review again — one more follow-up commit (`554257f`) addressing two items.

**Latency budget on the leading ack** (independent audit gap). `test_prompt_emits_leading_ack` now captures wall-clock latency between request publish and the first frame's arrival, asserts `< 500 ms`. The §6.4 rationale ("ack precedes the §6.6 warm-up gap") isn't enforced by wire ordering alone — a regression that wedged synchronous work between decode and the ack `await` would still produce ack-first ordering but defeat the spec invariant. 500 ms is generous (healthy local-NATS round-trip is ~5 ms here, so ~100× headroom) so CI flake risk is negligible. Evidence in `tests/_evidence/.../ack-latency.json`.

**Your minor observation** (`test_malformed_prompt_no_ack_before_400` assertion readability). Pulled the De Morgan into a named `is_leading_ack` predicate and extended the comment to make explicit that the `len(frames) == 2` + error-code-header pair already pins the §6.4 invariant — the explicit ack check is documentation-grade belt-and-braces.

Pre-push gate green on both SDKs: ruff (no-cache) ✓ · ruff format ✓ · mypy (no-incremental) ✓ · pytest 40 ✓ (agent-sdk) · pytest 212 ✓ (client-sdk).

renerocksai:
Alright, as signaled in OOB comms, I'm going to merge this on the Python side. 

#100 also provides the §6.4 leading-ack fix, for both TypeScript and Python SDKs. After this PR has been merged, @M64GitHub amends #100 to drop all `agent-sdk/python/**` changes and keeps only the TypeScript side (`agent-sdk/typescript/**` + TS CHANGELOG). Zero TS overlap from this PR, so #100's TS work is unaffected.

PR #113, opened 2026-05-18
agents/claude-code README: drop --dangerously-load-development-channels

M64GitHub:
## Summary
- Claude Code v2.1.143+ no longer requires `--dangerously-load-development-channels` — `claude` on its own picks up the installed plugin, so the quick-setup command drops to just `claude`.
- Tighten the Bun prerequisite to call out that `bun` must be on `PATH` after install, in cleaner English than the original draft.

## Test plan
- [ ] Eyeball the rendered README on GitHub
- [ ] Verify on a fresh box: install plugin via `/plugin install nats-channel@synadia-plugins`, then run plain `claude` and confirm the MCP server starts

### DIFF
--- a/agents/claude-code/README.md
+++ b/agents/claude-code/README.md
@@ -16,7 +16,7 @@ message - the protocol's uniform end-of-stream signal.
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


M64GitHub:
We still need the flag for messaging.

Issue #121, opened 2026-05-20
Pi NATS channel: expose reconnect/backoff config


randybias:
## Problem

`@synadia-ai/nats-pi-channel` calls `connect(opts)` with opts derived from the NATS context file (`contextToConnectOpts(natsCtx)`). The resulting connection uses nats.js defaults, which give up after ~10 reconnect attempts (~20s).

In multi-agent setups where the operator's host suspends (laptop sleep) or the NATS server is intermittently reachable, the 10-attempt budget burns through during the outage and the Pi-side NATS channel goes permanently dead — even after host network recovers — until the Pi process is restarted.

## Reproduction

1. Start a Pi agent using the channel.
2. Close laptop lid for >20s.
3. Re-open laptop. Host network resumes; Pi process is still alive.
4. Pi-side NATS channel does not reconnect; agent appears off-bus indefinitely.

## Request

Expose reconnect/backoff configuration for the extension, e.g. via the NATS context file, an environment variable, or extension options. A minimal API would be:

```jsonc
// context file
{
  "url": "...",
  "token": "...",
  "reconnect": {
    "maxAttempts": -1,
    "initialWaitMs": 2000,
    "maxWaitMs": 60000,
    "jitterMs": 200
  }
}
```

This would let consumers (we're using it in THAOS — github.com/randybias/thaos) pick a more aggressive policy without having to fork or runtime-patch.

## Workaround we're using

While this is open: wrapping `@synadia-ai/nats-pi-channel` in a thin extension that runtime-patches `connect` from `@nats-io/transport-node` to inject our resilience config before the upstream extension runs. We'd rather drop the patch once a config knob exists.

Happy to send a PR if the API shape is agreed.

derekcollison:
Thanks for the report @randybias - and great to see you here!

Will get this resolved.

M64GitHub:
Thanks for the great write-up @randybias - repro is spot-on and we just hit the same thing locally (laptop sleep + restart of a local nats-server).

Quick note on where we're landing:

- Defaults, not just knobs. To me an agent runtime should keep trying to reach the bus by default, so we're going to bake maxReconnectAttempts:
-1 (plus waitOnFirstConnect: true, so pi also recovers if it's started while the server is down) into the extension itself rather than asking every deployer to set it. A config block can come later as an advancedoverride.
[...]

PR #123, opened 2026-05-26
client-sdk/typescript: add withAgentReconnectDefaults helper (#121)

M64GitHub:
## Summary

- Adds `withAgentReconnectDefaults(opts)` and `AGENT_RECONNECT_DEFAULTS` to `@synadia-ai/agents`. Pure transform that fills resilient reconnect defaults into a caller's `NodeConnectionOptions`, preserving any explicit override (including `0` / `false`).
- Defaults: `maxReconnectAttempts: -1` (never give up), `reconnectTimeWait: 2000`, `reconnectJitter: 200`, `waitOnFirstConnect: true` (retry through startup too).
- CHANGELOG entry under `[Unreleased]`. No version bump in this PR.

## Why

[#121](https://github.com/synadia-ai/synadia-agents/issues/121) — `@nats-io/transport-node`'s defaults (10 attempts / ~20s) give up too fast for an agent process that's supposed to outlive laptop sleeps and intermittent broker reachability. While reproducing the issue locally we also noticed that starting an agent while the server is down fails immediately with no retry; `waitOnFirstConnect: true` is the second half of the fix.

Helper is opt-in — existing callers that don't wrap their opts keep upstream behavior.

## Scope discipline

This PR is the SDK helper only. The four downstream sites — `agents/pi`, `agents/openclaw`, `agents/claude-code`, `examples/pi-headless` — adopt the helper and add `case "close"` to their status loops in a separate follow-up PR. We will **not** publish to npm until that PR has been verified end-to-end locally (dev-cycle pattern with `./devtools/devmode.sh`).

Deferred to a later PR: context-file `reconnect` block + env-var overrides for deployer-level tuning.

## Test plan

- [x] `bun run check` (typecheck + lint + format + 290 tests including 10 new) green locally.
- [x] Unit tests cover: empty input fills all defaults; explicit `0` / `false` preserved; output is a fresh object (no input mutation); unrelated fields (`name`, `servers`, `token`, `inboxPrefix`) pass through; `AGENT_RECONNECT_DEFAULTS` is frozen.
- [ ] CI matrix (Node 20/22/24, Bun 1.2/latest).

### DIFF
--- a/client-sdk/typescript/CHANGELOG.md
+++ b/client-sdk/typescript/CHANGELOG.md
@@ -13,6 +13,29 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Added
+
+- `withAgentReconnectDefaults(opts)` and `AGENT_RECONNECT_DEFAULTS` —
+  opinionated reconnect defaults for agent runtimes:
+  `maxReconnectAttempts: -1`, `reconnectTimeWait: 2000`,
+  `reconnectJitter: 200`, `waitOnFirstConnect: true`. Pure transform
+  that fills in only the fields the caller left undefined, so explicit
+  `0` / `false` overrides survive untouched. See
+  [#121](https://github.com/synadia-ai/synadia-agents/issues/121).
+
+### Behavior notes
+
+- The new helper is opt-in. Existing callers that go straight to
+  `connect()` without wrapping their options keep the upstream
+  nats.js defaults (~20s reconnect budget, then `close`).
+- `waitOnFirstConnect: true` means `connect()` retries through a
+  server that is unreachable at startup. Misconfigured URLs or TLS
+  cert paths now manifest as a stuck "reconnecting…" UI rather than
+  an immediate throw — desirable for agent runtimes, surprising for
+  short-lived clients. Callers who want the throw-fast behavior
+  should pass `waitOnFirstConnect: false` explicitly (the helper
+  preserves it).
+
 ## [0.5.1] - 2026-05-11
 
 ### Changed
--- a/client-sdk/typescript/README.md
+++ b/client-sdk/typescript/README.md
@@ -73,13 +73,30 @@ Both error types extend `ValidationError` → `NatsAgentError`. See [Error handl
 
 ## What's in the box
 
-| API                                                              | Purpose                                                     |
-| ---------------------------------------------------------------- | ----------------------------------------------------------- |
-| `new Agents({ nc, ... })`                                        | Construct from a caller-owned `NatsConnection`.             |
-| `agents.discover({filter?, timeoutMs?})`                         | Return a live `Agent[]`; auto subscribe-before-ping (§8.5). |
-| `agent.prompt(text, {attachments, signal, inactivityTimeoutMs})` | Return a `PromptStream`.                                    |
-| `agents.liveness(id)` / `onHeartbeat(id, cb)` / `ping(id)`       | Heartbeat tracking and on-demand ping.                      |
-| `agents.close()`                                                 | Tear down SDK state; aborts all in-flight streams.          |
+| API                                                              | Purpose                                                                              |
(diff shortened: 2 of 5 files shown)


M64GitHub:
@claude - please review

M64GitHub:
## Review — `withAgentReconnectDefaults` helper

**Verdict: safe to merge.** Pure-additive, opt-in SDK helper with **zero runtime behavior change anywhere in the repo** (no caller is wired to it — adoption is the separate follow-up). Logic is correct, every factual claim in the comments/CHANGELOG checks out against the installed `@nats-io/*@3.4.0` source, and the full local CI-equivalent suite is green. Reviewed 100% of the diff (205/205 additions across all 4 files).

### Correctness (core logic)

```ts
const out = { ...opts };
for (const k of Object.keys(AGENT_RECONNECT_DEFAULTS)) {
  if (out[k] === undefined) out[k] = AGENT_RECONNECT_DEFAULTS[k];
}
```

- **`=== undefined` (not falsy)** is the load-bearing detail and it's right: explicit `0` (`maxReconnectAttempts: 0` = no reconnect) and `false` (`waitOnFirstConnect: false`) survive. A `!out[k]` check would have silently clobbered both. Directly tested.
- **No input mutation** — shallow copy, only top-level scalar keys are added, nested objects untouched; the guarantee holds for what the function does. Tested.
- Frozen constant is only *read*; values are primitives → no shared-mutable-state aliasing. The `as Record<string, unknown>` write cast is sound (all 4 keys are real `NodeConnectionOptions` fields with matching primitive types). No prototype-pollution surface (iterates a fixed frozen object, not user input).

### Every factual claim verified against `@nats-io/*@3.4.0`

| Claim in comments / CHANGELOG | Verified in source |
|---|---|
| upstream defaults: 10 / 2000 / 100 / 1000 / `false` | `options.js`: `DEFAULT_MAX_RECONNECT_ATTEMPTS=10`, `DEFAULT_RECONNECT_TIME_WAIT=2000`, `DEFAULT_JITTER=100`, `DEFAULT_JITTER_TLS=1000`, `waitOnFirstConnect:false` ✓ |
| `-1` = never give up | `core.d.ts` doc + `dialLoop` guard `if (mra !== -1 && reconnects >= mra)` ✓ |
| `waitOnFirstConnect:true` retries through startup | `dialLoop` catch: `if (!connectedOnce) { if (waitOnFirstConnect) continue; ... }` ✓ |
| residual `close` on repeated auth errors (we don't override `ignoreAuthErrorAbort`) | `handleAuthError` sets `abortReconnect=true` after 2 consecutive auth errors → top-of-loop `throw` → `close` ✓ |

### Build verification (local, == CI)

`typecheck` ✓ · `eslint` ✓ · prettier ✓ · **290 passed / 2 skipped** (the 2 skips are the env-gated websocket demo) · all **10 new tests** pass.

### Non-blocking notes (none gate the merge)

- **Downstream risk lives in the adoption PR, not here.** `waitOnFirstConnect:true` turns a misconfigured URL/cert from a fast throw into a stuck "reconnecting…", and callers then need `case "close"` handling — both documented well in the CHANGELOG "Behavior notes" and the source doc-comment. Worth confirming during the adoption PR's end-to-end run: whether a *first-connect handshake* auth rejection (vs. a post-establishment one) also hits the 2-error abort, or whether `waitOnFirstConnect`'s `continue` lets it retry the handshake indefinitely. Upstream nuance this PR neither introduces nor worsens.
- **Minor ergonomics:** `withAgentReconnectDefaults(opts)` requires an arg while `connect(opts?)` doesn't. Runtime-safe (`{...undefined}` → `{}`), and every in-repo caller builds an opts object anyway — could default to `= {}` for parity. Optional.
- **Doc ripple:** the new public export isn't in `client-sdk/typescript/README.md`'s API surface. CHANGELOG is updated (good); a README line would satisfy the ripple-check but reasonably belongs with the adoption PR.

---
🤖 Analysis generated with [Claude Code](https://claude.com/claude-code)

PR #124, opened 2026-05-26
agents/pi, openclaw, pi-headless: adopt resilient reconnect defaults + close handler (#121)

M64GitHub:
> [!WARNING]
> **Dev-cycle PR — do not merge until SDK is published.**
>
> Stacked on [#123](https://github.com/synadia-ai/synadia-agents/pull/123). Base is the SDK branch, not `main`, so the harness changes can typecheck against the new `withAgentReconnectDefaults` helper via the existing `file:` links. After #123 merges and a patched SDK is published to npm, this PR's base auto-retargets to `main` and a follow-up commit confirms the `^semver` pins on `package.json` are pointing at the new version (since main keeps `file:` by convention, no actual flip is needed for these three packages — the file: links keep pulling from local source until publish).
>
> **Publish gate:** the user has set the rule that no npm publish happens until this PR has been verified end-to-end against the manual scenarios in the test plan below. Reviewer bot may run; do not merge.

## Summary

- Adopts `withAgentReconnectDefaults` from `@synadia-ai/agents` (added in #123) at the three `connect()` call sites that aren't pinned by the marketplace: `agents/pi/extensions/nats-channel.ts`, `agents/openclaw/src/nats/connection.ts`, `examples/pi-headless/src/index.ts`.
- Adds `case "close"` to each status loop so the UI / log reflects an actually-dead connection instead of lying with "reconnecting…".
- Free fix while in `openclaw/src/nats/connection.ts`: replaces `(s as unknown as Record<string, unknown>).data` with the typed `s.server` (reconnect) and `s.error.message` (error). It had been logging `undefined` for the reconnect server name since the file was written.

`agents/claude-code` is listed in `devtools/.devmodeignore` because the marketplace install ships only that subtree and can't resolve `file:` links. Its adoption comes in a separate PR after the SDK is published.

## Why

[#121](https://github.com/synadia-ai/synadia-agents/issues/121). nats.js defaults give up after ~20s of reconnect attempts, so any operator-side outage longer than that window — laptop sleep, intermittent home network, broker maintenance — leaves the agent permanently off-bus until the process restarts. Plus the related "started while server down → immediate fail" path that `waitOnFirstConnect: false` causes.

## Test plan

- [x] `bun run build` in `client-sdk/typescript/` (the harnesses resolve the new helper via the existing `file:` links).
- [x] pi: `bunx tsc --noEmit extensions/nats-channel.ts` (no `tsc` script — bare-tsc clean).
- [x] openclaw: `bun run build` + `bun run test` (18 tests pass).
- [x] pi-headless: `bun run typecheck` (clean).
- [x] pi `bun test/smoke.mjs` regression check: 6 tests pass / 7 fail identically on `main` and this branch — the 7 failures are pre-existing on `main` and unrelated to this change.
- [ ] **Manual end-to-end verification** (the publish gate):
  1. Start nats-server locally; start agent (pi / openclaw / pi-headless). Stop nats-server. Wait ≥30s (past the old give-up window). Restart nats-server. Agent reconnects automatically. (Without this fix it would stay "reconnecting…" forever.)
  2. With nats-server stopped, start agent. Expect "reconnecting…" rather than immediate fail. Start nats-server. Agent connects.
  3. Trigger a terminal `close` (e.g. restart broker with auth pi can't satisfy → two repeated auth errors). UI / log reports "disconnected — agent is off-bus until restart" instead of the stuck "reconnecting…" lie.


M64GitHub:
## Review — adopt reconnect defaults + `close` handler

**Verdict: the code is correct — no introduced bugs.** All three harness changes do exactly what they claim, and I verified the load-bearing nats.js assumption against the installed source. Not mergeable *right now*, but only because of this PR's own intentional gates (stacked on #123; the `[!WARNING]` blocks merge until the SDK is published + manual E2E passes) — not because of any defect.

### Decisive question — does `case "close"` actually fire?

This is the whole point of the PR and easy to get wrong: in many NATS clients `close` is *not* a status event — the iterator just ends. Traced `close(err)` in `protocol.js@3.4.0`:

```js
for (const qi of this.listeners) {      // same queues nc.status() hands out
  qi.push({ type: "close" });           // ← close IS dispatched to consumers
  qi.stop();                            // ← then the iterator ends
}
```

`CloseStatus = { type: "close" }` is a real member of the `Status` union and is pushed to every status listener **before** the iterator stops. All three `case "close"` branches genuinely fire at runtime. The feature works.

### Per-file correctness

- **openclaw** — `(s as unknown as Record<string,unknown>).data` → `s.server` / `s.error.message` is a **genuine latent-bug fix**: there is no `data` field on any `Status` variant, so the old code logged `undefined` for the reconnect server name. New fields are type-correct (`DisconnectStatus`/`ReconnectStatus` carry `server: string`; `ServerErrorStatus` carries `error: Error`). Switch well-formed.
- **pi** — purely additive (import + `close` case + connect wrap). `disconnect`/`reconnect`/`error` already used `s.server`/`s.error.message` on the base, and `"warning"` severity is already used elsewhere in the file (line 456), so the new `notify(..., "warning")` is valid. No tsconfig (host-compiled) → bare-tsc is the right check.
- **pi-headless** — `s.server` / `s.error.message` / `s.type === "close"` all valid; comment sits correctly between `else if` arms.
- **All three** — `connect(withAgentReconnectDefaults(opts))` is at the right call site, and **none of the three set any reconnect option themselves**, so the helper's defaults apply with zero override interaction. pi sets `opts.name` before wrapping; it survives the spread.

### Verification (reproduced locally on the branch)

| Check | PR claim | Result |
|---|---|---|
| pi bare-tsc | clean | exit 0 ✓ |
| pi-headless `bun run typecheck` | clean | exit 0 ✓ |
| openclaw `bun run build` | ok | success ✓ |
| openclaw `bun run test` | 18 pass | 18 pass / 9 skip ✓ |

All three resolve a SDK build that actually exports `withAgentReconnectDefaults` (openclaw via symlink to live source; pi & pi-headless via refreshed `file:` copies).

### Merge-gating (process, not defects)

- **Stacked on #123** (base is the SDK branch). Correct order: merge #123 → GitHub auto-retargets #124 to `main` → then #124. The PR's own `[!WARNING]` already gates this.
- **No CI runs on this PR** ("no checks reported") — by design: the SDK workflows skip `agents/**` and `examples/**` paths. Local verification (above) + the `claude` reviewer bot are the gates; good that the manual run is green.
- **Publish gate** is the one open box: the 3 manual end-to-end scenarios (reconnect after >30s outage; start-while-server-down; terminal `close` on repeated auth error). Can't run those without a live nats-server, but the code paths behind all three are verified correct.

### Minor, non-blocking

1. **`close` also fires on intentional shutdown.** All three call `nc.drain()` on shutdown (pi:647, openclaw:64, pi-headless:98); drain → close emits `{type:"close"}`, so a clean shutdown logs/notifies "connection closed — agent is off-bus until restart" — slightly misleading noise. Documented behavior (the #123 doc-comment lists drain/close as a `close` path); optional to gate behind a "shutting down" flag.
2. **`file:` links live on `main`** for all three packages (pre-existing — this PR changes no `package.json`), including the published `pi-headless`. Not introduced here; semver pinning is the job of the post-publish follow-up (alongside `claude-code` adoption), which the PR body acknowledges.

**Bottom line:** correct, well-scoped, verified green. No introduced bugs. Merge once #123 lands and the manual E2E + publish gate is satisfied.

---
🤖 Analysis generated with [Claude Code](https://claude.com/claude-code)

PR #129, opened 2026-05-26
Add DeerFlow NATS channel

renerocksai:
## Summary

Adds an external DeerFlow channel for the Synadia Agent Protocol, packaged as `synadia-ai-nats-deerflow-channel` with the `deerflow-nats-channel` CLI.

- Registers a NATS agent service on `agents.{verb}.{token}.{owner}.{session}` and bridges prompt streams to DeerFlow Gateway.
- Adds config loading from CLI flags, environment variables, and TOML config files, plus a `doctor` command for operator diagnostics.
- Supports DeerFlow Gateway authentication/CSRF with automatic local login via username/password, while keeping manual cookie/CSRF config as a debug fallback.
- Creates DeerFlow Gateway threads before streaming so `/runs/stream` does not fail on missing sessions.
- Maps DeerFlow `ask_clarification` tool messages into protocol `query` chunks.
- Parses DeerFlow/LangGraph SSE streams path-aware so assistant text, tool activity, and internal metadata do not get mixed together.
- Adds user-facing install/configuration docs and a tag-triggered PyPI release workflow for the package.

## Stream mapping notes

The adapter intentionally does not recursively mine arbitrary strings from DeerFlow/LangGraph SSE payloads. It now:

- emits `response` only from explicit assistant/AI message content;
- emits DeerFlow tool calls/results as `status` chunks;
- emits `ask_clarification` as protocol `query` chunks;
- ignores provider/routing/metadata artifacts such as `openai`, `tool_calls`, `AIMessageChunk`, `stop`, and `branch:to:model`.

## Local verification

Run from `agents/deerflow`:

```bash
uv run pytest -q
uv run ruff check .
uv run ruff format --check .
uv run mypy src tests
uv build
```

Result:

```text
28 passed
ruff clean
format clean
mypy clean
sdist/wheel built successfully
```

Also smoke-tested against a local NATS server + DeerFlow Gateway on the M3 Max using:

```bash
nats req agents.prompt.df.rene.deerflow \
  "Give me a two sentence summary of what DeerFlow is." \
  --wait-for-empty \
  --reply-timeout 30s \
  --timeout 240s
```

Observed protocol behavior:

- leading `status` ack;
- streamed `response` chunks with DeerFlow answer text;
- zero-byte terminator at stream end;
- no LangGraph/provider metadata leaked as response text.

## Notes for reviewers

This PR treats DeerFlow as an HTTP Gateway dependency and does not modify/fork DeerFlow itself. The NATS bridge lives entirely under `agents/deerflow/`, with one small Python agent-service completion-error fix used by the new host tests.

### DIFF
--- a/.github/workflows/release-python-deerflow.yml
+++ b/.github/workflows/release-python-deerflow.yml
@@ -1,0 +1,97 @@
+name: Release — agents/deerflow
+
+# Publishes synadia-ai-nats-deerflow-channel to PyPI when a tag matching
+# `python-deerflow-v*` is pushed. The tag prefix keeps this package's
+# releases distinct from the monorepo's client-sdk (`python-v*`) and
+# agent-service (`python-agent-service-v*`) releases. Verification runs
+# before publishing and the publish job uses PyPI trusted publishing.
+#
+# Before the first release, register a pending publisher on PyPI for:
+#   project: synadia-ai-nats-deerflow-channel
+#   owner: synadia-ai
+#   repo: synadia-agents
+#   workflow: release-python-deerflow.yml
+#   environment: pypi
+
+on:
+  push:
+    tags:
+      - "python-deerflow-v*"
+
+concurrency:
+  group: release-python-deerflow-${{ github.ref }}
+  cancel-in-progress: false
+
+defaults:
+  run:
+    working-directory: agents/deerflow
+
+jobs:
+  verify:
+    name: Verify (py${{ matrix.python-version }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        python-version: ["3.11", "3.12", "3.13"]
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: astral-sh/setup-uv@v4
+        with:
+          enable-cache: true
+          cache-dependency-glob: "agents/deerflow/uv.lock"
+          python-version: ${{ matrix.python-version }}
+
+      - run: uv sync --all-extras
+      - run: uv run ruff check .
+      - run: uv run ruff format --check .
+      - run: uv run mypy src tests
+      - run: uv run pytest -v
+
+  publish:
+    name: Publish to PyPI
+    needs: verify
+    runs-on: ubuntu-latest
+    environment: pypi
+    permissions:
+      contents: write  # softprops/action-gh-release creates the tag's GitHub Release
+      id-token: write  # required for PyPI trusted publishing (OIDC)
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: astral-sh/setup-uv@v4
+        with:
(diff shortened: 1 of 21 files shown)


renerocksai:
Thanks for the careful pass — all seven findings were valid and are addressed in `a6f5a9c`.

What changed:

- Persisted an owned `httpx.AsyncClient` inside `DeerFlowGatewayClient` when no client is injected, added `aclose()`, and close it from both the runner helper and host prompt handler so login cookies/CSRF state survive clarification follow-up rounds without leaking sockets.
- Treated `POST /api/threads` `409 Conflict` as success so existing DeerFlow threads remain idempotent across follow-ups.
- Added a depth guard to `_state_message_payloads` to avoid `RecursionError` on pathological nested LangGraph state blobs.
- Reworked SSE semantic extraction so one SSE frame can emit multiple real messages instead of dropping the second message; metadata still stays filtered.
- Updated `doctor` to validate the same NATS target that `start` uses: explicit `NATS_URL`, explicit `NATS_CONTEXT`, or the current NATS CLI context fallback.
- Changed max-payload rejection in the agent service from protocol error code `413` to protocol-defined client error `400`.
- Removed the `--deerflow-password` CLI flag so passwords are not exposed via process listings; password remains supported through `DEERFLOW_PASSWORD`/TOML and is redacted in doctor output.

Verification run locally:

- `agents/deerflow`: `34 passed`, `ruff check`, `ruff format --check`, `mypy src tests`, `uv build`
- `agent-sdk/python`: `42 passed`, `ruff check`, `ruff format --check`, `mypy src tests`
- `git diff --check` clean before commit

CI on the pushed branch is green on py3.11/3.12/3.13 plus the automatic reviewer job.

renerocksai:
@claude please review again

renerocksai:
Handled the remaining `fake_deerflow_runner` item in `22f803e`.

- Moved the deterministic fake runner out of the production `runner.py` module and into `synadia_ai.nats_deerflow_channel.testing`.
- Updated host tests to import it from the testing-oriented module.
- `runner.py` no longer exposes `fake_deerflow_runner` as part of the production runner surface.

Local verification:

- `agents/deerflow`: `34 passed`
- `ruff check .`: passed
- `ruff format --check .`: passed
- `mypy src tests`: passed
- `uv build`: passed
- `git diff --check`: clean before commit

CI on the pushed branch is green for py3.11/3.12/3.13.

On the PyPI environment policy item: confirmed it is still an out-of-band repo setting rather than a code change. Current `pypi` environment policies include `python-agent-service-v*` and `python-v*`, but not `python-deerflow-v*`. The workflow itself is correct; before the first DeerFlow release tag, a repo admin should add a tag deployment policy named `python-deerflow-v*` to the `pypi` GitHub Environment.

renerocksai:
@claude please check once more

PR #137, opened 2026-06-03
fix: align Flue channel with conformance checklist

renerocksai:
## Summary

- Set Flue-hosted agents' protocol `session` metadata to the advertised instance name.
- Let `AgentService` map handler-raised `ProtocolError`s to `400` service errors while preserving `500` for ordinary handler failures.
- Reject decoded-but-unsupported Flue attachments as `ProtocolError` and extend protocol smoke coverage to validate metadata, endpoint shape, queue groups, `attachments_ok=false`, `max_payload`, status endpoint, prompt stream, and attachment rejection.

## Verification

- `cd agent-sdk/typescript && npm run typecheck && npm run test`
- `cd agents/flue && bun run typecheck && bun test && bun run smoke:protocol`
- `cd agents/flue && FLUE_BASE_URL=http://127.0.0.1:3583 FLUE_AGENT=echo FLUE_INSTANCE=conformance FLUE_SESSION=conformance SMOKE_PROMPT=conformance SMOKE_EXPECTS=echo:conformance bun run smoke:real-flue`
- `git diff --check`

Not pushed yet by request.

### DIFF
--- a/agent-sdk/typescript/CHANGELOG.md
+++ b/agent-sdk/typescript/CHANGELOG.md
@@ -6,6 +6,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Changed
+
+- `AgentService` now maps handler-raised `ProtocolError`s to
+  `Nats-Service-Error-Code: 400` responses while preserving the existing
+  `500` mapping for ordinary handler failures. This lets agent adapters reject
+  decoded-but-unsupported client input, such as attachments for an
+  `attachments_ok=false` endpoint, without misreporting the request as a server
+  failure.
+
 ## [0.5.2] - 2026-05-11
 
 ### Changed
--- a/agent-sdk/typescript/src/service.ts
+++ b/agent-sdk/typescript/src/service.ts
@@ -620,6 +620,8 @@ export class AgentService {
       try {
         const desc = err instanceof Error ? err.message : String(err);
         const isProtocolError =
+          // Cross-realm / duplicate-module guard: adapters may throw a
+          // ProtocolError class from another installed SDK copy.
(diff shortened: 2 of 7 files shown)


renerocksai:
Thanks for the review. One clarification: the original `"plain prompt"` case was not a JSON parse-time `ProtocolError` path — `decodeEnvelope` intentionally falls back to the §5.3 plain-text shorthand and routes raw bytes to the handler. So that part of the review read the decoder behavior incorrectly.

That said, the underlying request was fair: the handler-raised `ProtocolError` coverage should make the JSON-envelope path explicit instead of relying on shorthand behavior. I pushed follow-up commit `9012cfb` to make that unambiguous:

- added a dedicated integration test proving §5.3 plain-text shorthand reaches the handler;
- changed the handler-raised `ProtocolError` test to publish an explicit JSON envelope and assert the handler was actually called before the 400 frame is checked;
- added the cross-realm / duplicate-module comment for the `ProtocolError` name fallback;
- wrapped the Flue attachment-rejection raw subscription smoke path with a 5s timeout and unconditional unsubscribe so it cannot hang indefinitely.

Verification run locally:

```text
agent-sdk/typescript: npm run format:check && npm run typecheck && npm run test:integration
agents/flue: bun run typecheck && bun test && bun run smoke:protocol
```

Please re-review the follow-up commit.

renerocksai:
@synadia-claude-reviewer please review the follow-up commit `9012cfb`. The previous feedback is addressed above: JSON-envelope handler-raised `ProtocolError` coverage is now explicit, §5.3 plain-text shorthand coverage is preserved separately, and the Flue attachment-rejection smoke subscription has a timeout/unsubscribe guard.

renerocksai:
@claude Please re-review the follow-up commit.

PR #140, opened 2026-06-08
examples(agent-sdk/python): add the Python agent ladder (01-echo … 05-tools)

M64GitHub:
Brings the **Python** agent examples to parity with the TypeScript ladder. `agent-sdk/python/examples/` previously shipped only the reference agent; this adds the numbered ladder, mirroring `agent-sdk/typescript/examples/`, so the repo now has every agent example in **both** languages.

## What's added
- `01-echo.py`, `02-ollama.py`, `03-openrouter.py`, `04-combined.py` + the shared `llm.py` base — ported from `synadia-agents-labs/python/agent/` (the guarded versions).
- `05-tools.py` — **written from scratch** (no labs source): an Ollama `/api/chat` tool-calling agent whose `read_sensor` tool is backed by a NATS micro service. Python port of this repo's unique `05-tools.ts`.
- `agent-sdk/python/examples/README.md` documenting the ladder.

## Adapted to this repo's conventions (not a verbatim copy)
- **Connection** via the shared `_connect_cli.py` resolver (`--context`/`--url` + `$NATS_CONTEXT`/`$NATS_URL`), like every other Python example here.
- **Env-first identity + heartbeat**: `--owner` / `--session-name` / `--heartbeat-interval` each **default to** `NATS_AGENT_OWNER` / `NATS_AGENT_NAME` / `NATS_AGENT_HEARTBEAT_INTERVAL` (flags override). New shared `add_agent_identity_flags()` helper; `HEARTBEAT_INTERVAL=0` is treated as unset → default.
- `_reference_agent.py` flags now default to the same env vars too — **non-breaking** (its e2e test passes explicit flags, which still win).
- `httpx` added as an `examples` extra (CI runs `uv sync --all-extras`); only `02`–`05` need it.
- Backend config (`OLLAMA_*` / `OPENROUTER_*`) stays in env — the natural home for an API key.

## The strictness gotcha (Python analog of TS `exactOptionalPropertyTypes`)
Labs Python is only `py_compile` + `ruff E9,F` checked, but this repo runs **strict `mypy` over `examples/`**. The ports are typed to pass it (e.g. `json.loads(...)` results assigned through `dict[str, Any]` to avoid `warn_return_any`).

## Verification
`ruff check --no-cache`, `ruff format --check`, strict `mypy src tests examples`, and `pytest` (**42 passed** — incl. the retrofit-sensitive `_reference_agent` e2e) all clean. An echo round-trip smoke against a real `nats-server` confirms env-driven identity lands in the subject (`echo.<owner>.<session>`) and the `ack`→`response`→terminator wire shape.

Not included: client-side Python examples — they already exist here (`01-discover` … `06-chat`) and were untouched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ await service.stop();
 await nc.close();
 ```
 
-**Try it now:** [`agent-sdk/typescript/examples/01-echo.ts`](agent-sdk/typescript/examples/01-echo.ts) is this code packaged as a runnable script — `bun agent-sdk/typescript/examples/01-echo.ts` (with `$NATS_CONTEXT`, `$NATS_URL`, or localhost fallback).
+**Try it now:** [`agent-sdk/typescript/examples/01-echo.ts`](agent-sdk/typescript/examples/01-echo.ts) is this code packaged as a runnable script — `bun agent-sdk/typescript/examples/01-echo.ts` (with `$NATS_CONTEXT`, `$NATS_URL`, or localhost fallback). Both SDKs ship a parallel **agent ladder** (`01-echo` → `05-tools`: echo, Ollama, OpenRouter, combined, tool-calling) — TS in [`agent-sdk/typescript/examples/`](agent-sdk/typescript/examples/), Python in [`agent-sdk/python/examples/`](agent-sdk/python/examples/).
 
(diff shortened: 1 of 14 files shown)


M64GitHub:
@claude - please review

PR #143, opened 2026-06-08
examples(client-sdk/typescript): NATS_CONTEXT support in 01–05 + add README

M64GitHub:
Closes the TS client-examples gaps found while auditing env-var docs.

## 1. Harmonize connection handling (`01-discover` … `05-liveness`)
These five demos only read `$NATS_URL` (localhost fallback) — but `06-chat.ts`, **every** TS *agent* example, and **every** Python client example already do full `$NATS_CONTEXT → $NATS_URL → localhost` resolution. They were the only stragglers. Lifted the same block `06-chat.ts` uses, so they now:
- honor `$NATS_CONTEXT` (named NATS CLI context), and
- route URLs through `parseNatsUrl`, so `nats://TOKEN@host:port` credentials are no longer silently dropped.

No new logic invented — it's the established repo idiom (matches the labs versions verbatim).

## 2. Add `client-sdk/typescript/examples/README.md`
It was the **only example directory without a README** (Python client has one; both agent dirs have one). Documents the demos in a table and the connection env vars (`NATS_CONTEXT` / `NATS_URL`) — clients don't register agents, so the `NATS_AGENT_*` vars don't apply here, which the README notes.

## Verification
`bun run typecheck`, `lint`, and `format:check` all pass. Examples-only; no SDK surface change.


M64GitHub:
@claude - please review

PR #148, opened 2026-06-10
feat: add OpenCode NATS channel

renerocksai:
## Summary

This PR adds `@synadia-ai/opencode-nats-channel`: a Bun/TypeScript OpenCode integration that exposes OpenCode over the Synadia Agent Protocol for NATS.

It gives OpenCode users a protocol-compliant NATS channel with discovery, status, prompt streaming, and permission-query bridging while preserving OpenCode's own session and permission model.

## What changed

- Adds the new `agents/opencode` package with:
  - CLI entry points for managed and attached OpenCode operation
  - `AgentService`-backed Synadia protocol registration
  - OpenCode HTTP/SSE client wrappers
  - OpenCode plugin export at `@synadia-ai/opencode-nats-channel/opencode-plugin`
  - config parsing with clear CLI/env/file/default precedence
  - `doctor`, plugin install/uninstall, and environment-template UX
- Adds protocol and runtime coverage for:
  - discovery/status/prompt metadata
  - prompt streaming and upstream event mapping
  - unsupported attachment rejection as protocol 400
  - handler/upstream failures as protocol 500
  - permission.asked → Synadia query → OpenCode permission reply
  - plugin prompt routing without a preconfigured `OPENCODE_SESSION_ID`
  - concurrent same-session plugin prompts
- Updates public repo surfaces so OpenCode appears consistently in agent listings and the example web UI metadata.

### Plugin-first path

The plugin mode is the main UX path in this branch. It lets an OpenCode runtime register a Synadia channel through the production package export:

```ts
@synadia-ai/opencode-nats-channel/opencode-plugin
```

The maintained plugin smokes now exercise that package/export path through a generated wrapper. They do **not** rely on the older spike core.

### Session handling

OpenCode requires real upstream session ids (`ses...`). The plugin prompt bridge now lazily creates/reuses a valid OpenCode session when `OPENCODE_SESSION_ID` is absent and validates that no prompt call falls back to an invalid placeholder like `default`.

### Concurrency safety

Same-session plugin prompts are serialized per real OpenCode session id. The active prompt queue cleanup is ownership-guarded so one prompt cannot delete another prompt's event queue. A regression test covers the failure case where one prompt finishes while another is still waiting for permission events.

## Validation

Run from `agents/opencode` unless noted otherwise:

- [x] `bun install --frozen-lockfile`
- [x] `bun run typecheck`
- [x] `bun test` — 54 tests / 0 failures / 179 expectations
- [x] `bun run smoke:protocol`
- [x] `bun run smoke:opencode-lifecycle`
- [x] `bun run smoke:opencode-plugin-lifecycle`
- [x] `bun run smoke:opencode-plugin-permission`
- [x] `bun scripts/production-plugin-smoke.ts`
- [x] `git diff --check origin/main...HEAD`
- [x] Public/internal-term and secret-shape scan across the changed diff

Representative production-plugin smoke evidence:

```text
exercised: @synadia-ai/opencode-nats-channel/opencode-plugin via generated wrapper import
noConfiguredOpenCodeSessionId: true
sessionCreateCalls: 1
promptSessionIds: [ses_prod_*]
permissionReplies: [{ requestID: per_prod_1, reply: always }]
response: production plugin prompt complete
final status: done
```

## Known limits

- Attachments are intentionally unsupported in this first adapter surface and are advertised as `attachments_ok=false`; non-empty attachment envelopes return protocol 400.
- Credentialed OpenCode runtime/permission smokes are separate from the deterministic local smoke ladder and require a scoped local env file. They were not required for this deterministic plugin/protocol PR gate.
- Follow-up work for wrapper-launcher UX and multi-session manager behavior is intentionally left out of this PR.

## Reviewer focus

The highest-value review areas are:

1. `agents/opencode/src/plugin/prompt.ts` — session creation/reuse, per-session serialization, and guarded prompt cleanup.
2. `agents/opencode/scripts/production-plugin-smoke.ts` — proof that maintained smokes exercise the production plugin package/export path.
3. `agents/opencode/src/service.ts` and `src/nats.ts` — `AgentService` option construction and protocol metadata.
4. README/config docs — whether the managed, attached, and plugin modes are understandable for users.


renerocksai:
@claude can you please review?

renerocksai:
@claude Please re-review the follow-up commit `7705ce0`.

I addressed the review findings in a normal follow-up commit:

- Added per-session prompt serialization to the managed/attached `SdkOpenCodeBridgeClient` path, matching the plugin path's concurrency guard.
- Added a regression proving concurrent managed/attached prompts for the same OpenCode session serialize instead of overlapping.
- Tightened OpenCode session id validation to require `ses_...` on both SDK and plugin paths.
- Restricted plugin event fallback routing so sessionless permission events are not delivered to the only active prompt.
- Added a regression for sessionless permission events in plugin mode.
- Added warning logs for plugin permission reply fallback transitions.
- Extracted the duplicate async event queue into a shared `AsyncQueue<T>`.
- Changed `allow` query replies to map to `always` and added test coverage.
- Removed the internal implementation spec and spike sources from the PR/package surface.
- Updated package metadata tests and verified `npm pack --dry-run` contains no `PLUGIN_FIRST_IMPLEMENTATION_SPEC.md`, `spikes/`, `docs/html`, or `docs/research` entries.
- Added a note about quoting TOML password values containing `#` or spaces.

Local verification after the fix:

```text
cd agents/opencode
bun run typecheck
bun test
# 56 pass / 0 fail / 186 expect() calls

bun run smoke:protocol
bun run smoke:opencode-plugin-lifecycle
bun run smoke:opencode-plugin-permission
# all pass; production plugin smokes still exercise @synadia-ai/opencode-nats-channel/opencode-plugin via generated wrapper import

git diff --check --cached
# pass before commit

gh pr diff 148 --name-only
# no internal artifact filenames for agents/opencode/spikes, PLUGIN_FIRST_IMPLEMENTATION_SPEC.md, docs/html, docs/research, or secrets/
```

renerocksai:
@claude Please re-review the follow-up commit `2019529`.

I checked the remaining minor observation from your re-review (`nc.close()` vs `nc.drain()`) and made the shutdown path consistent with the plugin path and sibling agent precedent:

- `agents/opencode/src/cli.ts`: CLI shutdown now drains the NATS connection after `service.stop()` and `client.close?.()`.
- `agents/opencode/scripts/protocol-smoke.ts`: smoke cleanup now drains both service/client NATS connections before stopping the disposable `nats-server`.
- `agents/opencode/scripts/opencode-permission-smoke.ts`: same cleanup change for the real permission smoke.

Local verification:

```text
cd agents/opencode
bun run typecheck
bun test
# 56 pass / 0 fail / 186 expect() calls

bun run smoke:protocol
# pass

git diff --check
# pass before commit
```

I also attempted `bun run smoke:opencode-permission`; it did not reach the changed cleanup path because this machine does not have the scoped credential file required by that smoke:

```text
scoped OpenCode test env file does not exist: /Users/rs/.config/synadia/opencode-runtime-smoke.env
```

renerocksai:
Follow-up verification: I also ran the credentialed real-runtime permission smoke with a real OpenRouter model after creating a scoped local test env file.

No secrets are printed by the smoke, and the env file is repo-external with only the two allowed keys.

Verification command:

```text
cd agents/opencode
bun run smoke:opencode-permission
```

Result: passed.

Relevant evidence from the smoke output:

```text
opencodeMode: managed
model: openrouter/openai/gpt-5.5
targetExistsBefore: true
targetExistsAfterDeny: true
permissionQueryPromptIncludesTarget: true
rejectedSeen: true
exit code: 0
```

So the credentialed path now also proves the real OpenCode-managed runtime emitted a permission query, the Synadia protocol query replied deny, the protected target remained present, and the stream recorded the rejection status.

PR #150, opened 2026-06-12
feat(examples): adopt SYNADIA_* identity env var scheme in agent-sdk TS examples

M64GitHub:
## What

Migrates the five numbered `agent-sdk/typescript/examples` to the canonical identity env var lookup chain being adopted across `agents/*`:

```
SYNADIA_<AGENT>_OWNER  >  SYNADIA_OWNER  >  NATS_AGENT_OWNER (legacy)  >  $USER  >  "anon"
SYNADIA_<AGENT>_NAME   >  SYNADIA_NAME   >  NATS_AGENT_NAME  (legacy)  >  "main"
```

`<AGENT>` is each example's agent token (`ECHO`, `OLLAMA`, `OPENROUTER`, `LLM`, `TOOLS`).

## Why

- The examples are what new agent authors copy-paste, so they should demonstrate the convention: a per-agent var (scopes an override to one agent on a shared host) above a fleet-wide var (set identity once for everything).
- The legacy `NATS_AGENT_OWNER` / `NATS_AGENT_NAME` vars **keep working** as lowest-priority aliases — `synadia-agents-labs` and the Python example ladder were built from these files and still document them. Nothing breaks.
- The `agent` (3rd) subject token stays fixed per example by design — no override mechanism is introduced (project decision, 2026-06-12).
- `NATS_AGENT_HEARTBEAT_INTERVAL` is deliberately untouched; heartbeat var naming is a separate, undecided question.

## Docs

- `examples/README.md`: env table rows for the new scheme (legacy noted as honored), run snippet uses `SYNADIA_OWNER`/`SYNADIA_NAME`, plus a note that the chain is the convention to copy.
- `client-sdk/typescript/examples/README.md`: cross-reference sentence updated.
- `agent-sdk/typescript/CHANGELOG.md`: entry under `[Unreleased]` (examples-only, no package API change).

## Verification

- `tsc --noEmit`, `eslint`, `prettier --check` all green.
- Live precedence smoke test against a local nats-server: per-agent beats fleet-wide beats legacy, producing `agents.prompt.echo.peragent.t1`, `...fleet.t2`, `...legacy.t3` respectively.

### DIFF
--- a/agent-sdk/typescript/CHANGELOG.md
+++ b/agent-sdk/typescript/CHANGELOG.md
@@ -8,6 +8,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### Changed
 
+- **Examples: identity env vars moved to the `SYNADIA_*` scheme.** The
+  numbered examples now resolve `owner`/`name` through the per-agent >
+  fleet-wide > legacy chain (`SYNADIA_<AGENT>_OWNER` > `SYNADIA_OWNER` >
+  `NATS_AGENT_OWNER`, same for `_NAME`), matching the env naming
+  convention being adopted across `agents/*`. The legacy
+  `NATS_AGENT_OWNER` / `NATS_AGENT_NAME` vars keep working as
+  lowest-priority aliases; no package API change.
 - `AgentService` now maps handler-raised `ProtocolError`s to
   `Nats-Service-Error-Code: 400` responses while preserving the existing
   `500` mapping for ordinary handler failures. This lets agent adapters reject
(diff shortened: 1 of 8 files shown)


M64GitHub:
@claude - please review

M64GitHub:
Thanks — both findings acknowledged:

- **Sanitization observation**: covered by the SDK rather than the examples. `AgentService` routes all three tokens through `AgentSubject.new(...)`, which calls `assertValidToken` per token (`client-sdk/typescript/src/subjects.ts:123-125`), so a malformed env value fails fast at startup with a role-tagged error instead of registering a broken subject. The examples intentionally don't sanitize: opencode sanitizes because it *derives* tokens from messy sources (directory paths, `$USER`), whereas here the values are explicit user input — failing loudly is the better teaching behavior than silently rewriting. Agreed it's worth a sentence in a future hardening/convention doc pass.
- **Hyphenated-name nit** (`my-agent` → `SYNADIA_MY_AGENT_OWNER`): agreed, deferred — will be settled in the convention write-up that accompanies the upcoming `agents/*` migration PRs, where hyphenated agent names (`open-agent`, `claude-code`) actually occur.

PR #151, opened 2026-06-12
feat(claude-code): make owner settable; adopt SYNADIA_* identity env vars

M64GitHub:
Second PR in the identity env var series (follows #150, the examples migration).

## What

The owner (4th subject token) in `agents.prompt.cc.<owner>.<name>` was hard-wired to sanitized `$USER` — claude-code was the only agent plugin where it could not be configured at all. It now resolves through the `SYNADIA_*` identity convention:

```
owner: SYNADIA_CLAUDE_CODE_OWNER > SYNADIA_OWNER > config "owner" > sanitized $USER > "unknown"
name:  SYNADIA_CLAUDE_CODE_NAME  > SYNADIA_NAME  > NATS_SESSION_NAME (legacy) > config "sessionName" > sanitized basename($CLAUDE_CWD) > "default"
```

- **Zero behavior change when no new vars are set** — the default path produces byte-identical subjects.
- **`NATS_SESSION_NAME` keeps working** as a legacy alias, one notch below the `SYNADIA_*` vars.
- Sanitization philosophy preserved: explicit values flow raw to `AgentSubject.new` (which rejects invalid tokens loudly via `assertValidToken`); derived values (`$USER`, CWD basename) stay sanitized.

## Also in this PR

- `/nats-channel:configure` learns `owner <name>` / `owner clear` subcommands and shows the owner override in the status view (the skill is the documented writer of `config.json`, so the new config field gets a writer too).
- README: documents the new identity vars **and** the previously undocumented `NATS_CONTEXT`, `NATS_URL`, and `CLAUDE_CWD` (README-drift found during the env var analysis).

## Verification

5-case smoke matrix against an isolated local nats-server, asserting the registered subject from server stderr:

| Case | Env / config | Registered subject |
| --- | --- | --- |
| per-agent wins over fleet + legacy | `SYNADIA_CLAUDE_CODE_OWNER=peragent` `SYNADIA_OWNER=fleet` `SYNADIA_CLAUDE_CODE_NAME=t1` `SYNADIA_NAME=x` `NATS_SESSION_NAME=y` | `agents.prompt.cc.peragent.t1` |
| fleet owner, legacy name | `SYNADIA_OWNER=fleet` `NATS_SESSION_NAME=t2` | `agents.prompt.cc.fleet.t2` |
| config owner honored | config `{"owner":"cfgowner"}` | `agents.prompt.cc.cfgowner.t3` |
[...]

PR #152, opened 2026-06-12
feat(pi): adopt SYNADIA_* identity env vars; owner env now beats config

M64GitHub:
Third PR in the identity env var series (#150 examples, #151 claude-code).

## What

pi's owner and session name now resolve through the `SYNADIA_*` identity convention:

```
owner: SYNADIA_PI_OWNER > SYNADIA_OWNER > NATS_PI_OWNER (legacy) > config "owner" > $USER > "unknown"
name:  SYNADIA_PI_NAME  > SYNADIA_NAME  > NATS_SESSION_NAME (legacy) > config "sessionName" > CWD basename > "pi"
```

## ⚠️ The one behavior change in the series

**Owner precedence flip: env vars now beat the config file.** Previously `config.owner` (in `~/.pi/agent/nats-channel.json`) won over `$NATS_PI_OWNER` — the inverse of every other plugin *and* of pi's own session-name handling. Now env > config uniformly. **Only setups that set both the config field and an owner env var to different values are affected**; everyone else sees identical behavior. Given pi's community install base this is called out prominently in a new `CHANGELOG.md` (pi had none; started one in Keep-a-Changelog format).

Legacy `NATS_PI_OWNER` / `NATS_SESSION_NAME` keep working indefinitely as lower-priority aliases.

## Changes

- `extensions/subject.ts`: `resolveOwner` is now variadic (sources highest-precedence first); the documented first-present-wins-then-sanitize semantics are preserved exactly (`s != null` mirrors `??`).
- `extensions/nats-channel.ts`: new resolution chains; `/nats-configure` learns `owner <name|clear>` and shows the owner in its status line.
- `test/owner.test.ts`: rewritten for the 5-source chain — including an explicit test pinning the env-beats-config flip. **9/9 pass.**
- `test/smoke.mjs`: identity env vars (`SYNADIA_*`, `NATS_PI_OWNER`) are now deleted at startup so overrides leaking in from the invoking shell can't skew the expected subject.
- `README.md`: env table, resolution-order section, command table; legacy vars documented as still honored.
- `CHANGELOG.md`: new file with the breaking-precedence callout.

## Verification

- `bun test test/owner.test.ts`: 9/9 pass.
- `bun test/smoke.mjs`: registration assertion passes — extension registers `agents.prompt.pi.m64.smoke-<pid>` with correct `metadata.owner`; running with `USER="Funky Owner"` registers `agents.prompt.pi.funky-owner.…` (sanitization through the real extension), and `SYNADIA_OWNER=leaked-from-shell` is hermetically ignored.
- **Pre-existing, out of scope:** the smoke fails 7/13 on *unmodified main* identically (heartbeat-cadence expectation drift — smoke expects `interval_s` 30, extension publishes 5 — plus prompt-flow mock drift, e.g. `pi.sendUserMessage was not called`). My change adds zero new failures. Worth a separate fix-the-smoke issue/PR.

### DIFF
--- a/agents/pi/CHANGELOG.md
+++ b/agents/pi/CHANGELOG.md
@@ -1,0 +1,32 @@
+# Changelog
+
+All notable changes to `@synadia-ai/nats-pi-channel` will be documented in
+this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Changed
+
+- **Identity env vars adopt the `SYNADIA_*` convention** shared across
+  `agents/*`. Owner: `SYNADIA_PI_OWNER` > `SYNADIA_OWNER` >
+  `NATS_PI_OWNER` (legacy) > config `owner` > `$USER` > `unknown`.
+  Session name: `SYNADIA_PI_NAME` > `SYNADIA_NAME` >
+  `NATS_SESSION_NAME` (legacy) > config `sessionName` > CWD basename.
+  The legacy vars keep working indefinitely as lower-priority aliases.
+- **BREAKING (owner precedence): env vars now beat the config file.**
+  Previously the `owner` field in `~/.pi/agent/nats-channel.json` won
+  over `$NATS_PI_OWNER`; now any owner env var wins over the config
+  field — uniform with flue, opencode, openclaw, open-agent, and pi's
+  own session-name handling. Only setups that set *both* the config
+  `owner` field *and* an owner env var to different values are
+  affected; everyone else sees no change.
+- `/nats-configure` learns `owner <name|clear>` and shows the owner
+  override in its status output.
+
+## [0.5.6] and earlier
+
+Changelog started 2026-06-12 (at package version 0.5.6); see the git
+history of `agents/pi/` for earlier changes.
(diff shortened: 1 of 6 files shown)


M64GitHub:
@claude - please review

M64GitHub:
Thanks — acknowledging the two optional follow-ups rather than addressing them here (both out of scope per the review):

1. **Stale smoke assertions** (`interval_s === 30` vs `HEARTBEAT_INTERVAL_S = 5`, plus the prompt-flow mock drift — 7/13 failing on unmodified main): deserves its own fix-the-smoke PR; assert against the actual constant rather than a literal.
2. **`/nats-configure` env-override echo**: agreed it's a nice UX lift — a parenthetical naming the active identity env var on the relevant status line, keeping the command a config editor.

Both noted for follow-up after the identity env var series completes.

M64GitHub:
@claude - please re-review

M64GitHub:
Recommendation addressed in the latest commit: env-sourced session names now go through `sanitizeSubjectToken` with the same first-present-wins-then-sanitize semantics as `resolveOwner` (winner coerced; empty-sanitizing winner falls back to `"pi"` without cascading — consistent with the owner behavior pinned in `owner.test.ts`).

One note on the suggested snippet in the review: it mixes `??` and `||` without parentheses, which TypeScript rejects — the committed form sanitizes the whole resolved winner instead, which also keeps `config.sessionName` (sanitized-on-write, idempotent re-sanitize) and the CWD basename on the identical path.

9/9 unit tests and the smoke registration baseline unchanged.

PR #153, opened 2026-06-12
feat(openclaw): adopt SYNADIA_* identity env vars; accept NATS_CREDS alias

M64GitHub:
Fourth PR in the identity env var series (#150 examples, #151 claude-code, #152 pi).

## What

openclaw's identity env overrides follow the `SYNADIA_*` convention:

```
owner:     SYNADIA_OPENCLAW_OWNER > SYNADIA_OWNER > NATS_OWNER (legacy) > NATS_ORG (legacy) > account config
agentName: SYNADIA_OPENCLAW_NAME  > SYNADIA_NAME  > NATS_AGENT_NAME (legacy) > account config
```

**Purely additive — zero behavior change for existing setups.** The legacy vars keep identical semantics; the new vars only sit above them. Also: `NATS_CREDS` is now accepted as a credentials alias (used only when `NATS_CREDENTIALS` is unset — the incumbent wins when both are set), matching flue and opencode.

## Changes

- `src/accounts.ts`: new `firstDefinedEnv` helper returns the winning var's *name* alongside the value, so the existing env-override log line attributes the value to the var that actually supplied it. The `listAccountIds` debug line now reports the resolved name var instead of hardcoding `NATS_AGENT_NAME`.
- `src/accounts.test.ts`: new "identity env overrides" suite (6 tests covering both chains, both legacy owner aliases, and both creds-alias directions), plus hermetic snapshot/restore of identity env vars for the existing suites — a leaked `SYNADIA_OWNER` or `NATS_CREDS` in the invoking shell can no longer skew expectations.
- `README.md`: env table documents the new vars, the convention, and the alias semantics.
- `CHANGELOG.md`: new file (Keep a Changelog format).

## Verification

Full vitest suite: **33/33 pass, zero skipped** — with the `openclaw` peer dep made resolvable (scratch-dir install + symlink; note `npm i --no-save openclaw` is a silent no-op against the empty-string optional peer range). `tsc --noEmit` error count unchanged vs main (11 lines, all pre-existing SDK-type drift in `channel.ts`/`gateway.ts` plus vitest types; none in `accounts.ts`).

## Observability gap (for a follow-up, not this PR)

The accounts suite **auto-skips when the peer dep is absent**, and no CI workflow covers `agents/openclaw` (or any `agents/*`) — the same gap that let the pi smoke drift. A follow-up `agents/*` CI PR is planned.

### DIFF
--- a/agents/openclaw/CHANGELOG.md
+++ b/agents/openclaw/CHANGELOG.md
@@ -16,8 +16,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   name: `SYNADIA_OPENCLAW_NAME` > `SYNADIA_NAME` > `NATS_AGENT_NAME`
   (legacy) > account config. Purely additive — the legacy vars keep
   working with identical behavior; the new vars only sit above them.
-- `NATS_CREDS` is accepted as an alias for `NATS_CREDENTIALS` (used
-  only when `NATS_CREDENTIALS` is unset), matching flue and opencode.
+- `NATS_CREDS` is accepted as an alias for `NATS_CREDENTIALS` — the
+  spelling flue and opencode also accept. Tie-break differs
+  deliberately: here the incumbent `NATS_CREDENTIALS` wins when both
+  are set (zero change for existing deployments), whereas flue/opencode
+  check `NATS_CREDS` first.
 - The env-override log line now names the variable that actually
   supplied the value (e.g. `SYNADIA_OWNER`), not just the legacy name.
 
--- a/agents/openclaw/CHANGELOG.md
+++ b/agents/openclaw/CHANGELOG.md
@@ -1,0 +1,27 @@
+# Changelog
+
+All notable changes to `@synadia-ai/nats-channel` (the OpenClaw plugin)
+will be documented in this file.
+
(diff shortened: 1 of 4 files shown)


M64GitHub:
@claude - please review

M64GitHub:
All four findings addressed in 0193774:

1. **NATS_CREDS precedence claim corrected** — you're right, flue/opencode check `NATS_CREDS` first. Only the accepted *spelling* is shared; openclaw's tie-break is deliberately inverted (incumbent `NATS_CREDENTIALS` wins) so existing deployments see zero change when both are set. Now documented in a code comment at the `credsEnv` site and corrected in the CHANGELOG.
2. **README table grouping** — `NATS_DESCRIPTION` moved above the identity groups; the name and owner runs are now uninterrupted.
3. **Quickstart** — inline `(or SYNADIA_OPENCLAW_NAME)` / `(or SYNADIA_OPENCLAW_OWNER)` notes added to the legacy exports.
4. **`loggedEnvOverrides`** — comment added stating the dedup affects logging only; the field assignment always precedes the dedup check.

vitest suite still 33/33 with the peer dep resolvable.

M64GitHub:
@claude - please re-review

PR #158, opened 2026-06-16
fix(client-sdk): decodeEnvelope follows §5.3 discrimination exactly

M64GitHub:
While diagnosing the pi smoke failures (for the upcoming agents/* CI PR), the smoke turned out to be catching a **real spec-compliance bug** in the TS SDK decoder — not stale assertions. This PR fixes the bug; a follow-up PR handles the genuinely-stale heartbeat assertion + CI.

## The bug

`decodeEnvelope` discriminated JSON-vs-plain-text by "try `JSON.parse` the whole payload; on **any** failure, treat it as plain text." That violates spec **§5.3** in two ways:

| Input | Spec §5.3 | Old TS | Python SDK | New TS |
| --- | --- | --- | --- | --- |
| `""` (zero-byte) | **400** ("MUST be rejected") | accepted as `{prompt:""}` | 400 | **400** ✓ |
| `{not json` (leading `{`, bad JSON) | **400** (committed to JSON) | accepted as `{prompt:"{not json"}` | 400 | **400** ✓ |
| `Hello` / `5` / `[1,2]` / `"x"` | plain text | plain text | plain text | plain text ✓ |
| `{"prompt":"hi"}` | envelope | envelope | envelope | envelope ✓ |

Both gaps were also live **TS↔Python wire divergences** — the Python SDK already implements §5.3 correctly (`looks_like_json`), so this aligns TS *to* Python rather than the reverse. No Python change required.

## The fix

Discrimination now keys off the **first non-whitespace byte being `{`** (new `startsWithJsonObject` helper, mirroring Python's `looks_like_json`):
- first non-ws byte `{` → MUST parse as JSON; parse failure → `ProtocolError` (400).
- anything else → plain-text shorthand (whole payload is the prompt).
- zero-byte → rejected up front.

The decode **signature is unchanged**, so pi and claude-code (which route envelope decode through this SDK) inherit the fix with no edit. Everything else is unchanged: plain-text shorthand for non-`{` payloads, leading-whitespace tolerance, and rejection of missing/non-string/empty `prompt` + invalid attachments.

## Scope note (`{"prompt":""}`)

There is a third, **softer** TS↔Python divergence: a JSON envelope with an empty-string prompt. TS rejects it at decode; Python allows it and defers to §5.4/server. This PR **does not touch** that behavior (TS keeps rejecting). Reconciling it toward Python's "allow at decode" would require adding empty-prompt 400 checks in 4 host paths (`AgentService` + pi + claude-code + openclaw) to avoid silently regressing the 400 those agents return today — a distinct, broader change worth its own PR. Flagged for follow-up.

## Verification

- `client-sdk`: **294 pass / 2 skip**, typecheck + lint clean. New unit cases: zero-byte → throw, `{`-led-malformed → throw, leading-whitespace discrimination.
- `agent-sdk` (consumes `decodeEnvelope`): **40 pass** — the §5.3 plain-text routing test still passes.
- Against the rebuilt SDK, the pi smoke's 5 cascade failures clear (11/13; the 2 remaining are the stale `interval_s` 5-vs-30 assertion, fixed in the follow-up).

### DIFF
--- a/client-sdk/typescript/CHANGELOG.md
+++ b/client-sdk/typescript/CHANGELOG.md
@@ -13,6 +13,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+- **`decodeEnvelope` now follows the §5.3 discrimination algorithm
+  exactly**, closing two wire-compatibility gaps with the spec and the
+  Python SDK:
+  - A **zero-byte request payload** is now rejected with a
+    `ProtocolError` (→ agent `400`). Previously it was silently accepted
+    as `{ prompt: "" }`. Spec §5.3: "A zero-byte request payload is
+    invalid and MUST be rejected with status `400`."
+  - A payload whose **first non-whitespace byte is `{` but which is not
+    well-formed JSON** (e.g. `{not json`) is now rejected as a malformed
+    envelope, instead of being mis-classified as a plain-text prompt
+    literally equal to the broken JSON. Discrimination now keys off the
+    leading `{` (matching the Python SDK's `looks_like_json`) rather than
+    "try `JSON.parse`, fall back to plain text on any failure".
+
+  Plain-text shorthand for payloads that don't lead with `{` (including
+  bare numbers, `[`, or quotes) is unchanged, as is rejection of
+  missing / non-string / empty `prompt` and invalid attachments. The pi
+  spec-compliance smoke (`agents/pi/test/smoke.mjs`) caught both gaps.
+
 ## [0.5.2] - 2026-05-26
 
 ### Added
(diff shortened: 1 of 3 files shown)


M64GitHub:
@claude - please review

M64GitHub:
Thanks! Added tests for the two whitespace edge cases (latest commit) — both worth pinning since they're exactly why discrimination is a hand-rolled ASCII byte check rather than `trimStart().startsWith("{")`:

- all-whitespace `"   "` → plain-text shorthand, verbatim
- `\f`-led (`\f{...}`) → `\f` is JS whitespace but **not** §5.3 whitespace, so it's the first non-ws byte → plain text, never mis-classified as JSON

Left the `utf8("")` / `new Uint8Array(0)` pair in the zero-byte test as-is — harmless, and it documents that both call shapes hit the guard. decode-envelope suite green.

PR #159, opened 2026-06-16
ci(agents): run pi + openclaw suites; fix stale pi heartbeat assertion

M64GitHub:
Stacked on #158 (the decode fix). **Base is `sdk/decode-envelope-5.3-compliance`** — GitHub will retarget this to `main` once #158 merges. The pi smoke only goes fully green with #158's decode fix present, hence the stacking.

This is the CI + smoke-staleness PR we'd queued. Diagnosing the pi smoke for it is what surfaced the #158 SDK bug; this PR handles the rest.

## What

**New workflow `.github/workflows/agents-typescript.yml`** — the TS agent harness suites were never wired into CI (and openclaw's couldn't run at all; see below). Two jobs, triggered on `agents/pi/**`, `agents/openclaw/**`, **and** `client-sdk/typescript/**` + `agent-sdk/typescript/**` (both agents link the SDKs via `file:`, so an SDK change that breaks an agent now surfaces here):

- **pi (bun):** build both SDKs → `test/owner.test.ts` (unit) + `test/smoke.mjs` (spec-compliance smoke; spawns its own nats-server). **13/13** with #158.
- **openclaw (node):** build both SDKs → install + peer → vitest. **33/33, zero skipped.**

## Two latent gaps that kept openclaw's tests from ever running

1. **vitest was never declared.** The `test` script is `vitest run`, but `vitest` wasn't in `devDependencies` — so `npm test` couldn't find a runner anywhere. Added `vitest@^3.2.0` (matching the SDKs).
2. **The optional `openclaw` peer is uninstallable by normal means.** Its empty-range optional-peer declaration makes npm treat it satisfied-by-absence: plain `npm install`, `npm install --no-save openclaw`, and even `npm install openclaw` all leave it absent — silently skipping the 15-test account-resolution suite. The workflow promotes it to a real dep with `npm pkg set "dependencies.openclaw=2026.6.5"` (ephemeral on the CI checkout) so the tree installs, then **asserts `openclaw/plugin-sdk/core` resolves** so a broken peer fails the job rather than hollowing it out.

## pi heartbeat staleness (the only genuinely-stale assertion)

The smoke asserted `interval_s === 30`, but the extension pins `HEARTBEAT_INTERVAL_S = 5` (the snappy first-party cadence). Per your ask that **both 5s and 30s pass**, I exported the constant from the extension and assert the advertised `interval_s` against it — the test now tracks the source of truth. Proven by flipping the constant to 30 locally (still 13/13), reverted to 5.

## Verification (local)

- pi smoke **13/13** (against #158's rebuilt SDK); flipped-to-30 also 13/13.
- openclaw **33/33, zero skipped**; `openclaw/plugin-sdk/core` resolves.
- Clean dry-run of the exact openclaw job steps from an empty `node_modules`.

### DIFF
--- a/.github/workflows/agents-typescript.yml
+++ b/.github/workflows/agents-typescript.yml
@@ -79,10 +79,18 @@ jobs:
       # dependency (ephemeral on the CI checkout) so the full tree installs,
       # then assert the subpath the tests import actually resolves — a broken
       # peer fails the job instead of hollowing it out.
-      - name: Install openclaw deps + peer
+      #
+      # Resolve the LATEST openclaw rather than pinning: this plugin's whole
+      # job is to track openclaw, so a breaking change to `plugin-sdk/core`
+      # SHOULD surface here (the resolve-assert guards presence; a broken API
+      # surfaces as a vitest failure). A pin silently rots — openclaw shipped
+      # 2026.6.5 -> 2026.6.6 within days of this workflow landing.
+      - name: Install openclaw deps + peer (latest)
         working-directory: agents/openclaw
         run: |
-          npm pkg set "dependencies.openclaw=2026.6.5"
+          OPENCLAW_VERSION="$(npm view openclaw version)"
+          echo "openclaw peer: $OPENCLAW_VERSION"
+          npm pkg set "dependencies.openclaw=${OPENCLAW_VERSION}"
           npm install
           node -e "require.resolve('openclaw/plugin-sdk/core')"
       - name: Tests (account resolution + protocol + context loader)
--- a/.github/workflows/agents-typescript.yml
+++ b/.github/workflows/agents-typescript.yml
@@ -1,0 +1,90 @@
+name: CI — agents (TypeScript)
+
+# Runs the test suites for the TypeScript agent harnesses, which depend on
+# the TS SDKs via `file:` links — so SDK changes trigger this too, catching
+# cross-subtree regressions (e.g. an envelope-decode change breaking the pi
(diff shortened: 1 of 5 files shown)


M64GitHub:
@claude - please review

M64GitHub:
Thanks — addressed the main flagged risk and confirmed the others.

**Adopted: dynamic openclaw version** (latest commit). Switched the peer step to resolve `npm view openclaw version` instead of the `2026.6.5` pin. Your concern proved itself immediately — `npm view` now returns **`2026.6.6`**, so the pin was already a release behind. Verified vitest stays **33/33** against the latest. Rationale baked into the workflow comment: this plugin's job is to track openclaw, so a breaking `plugin-sdk/core` change *should* fail here rather than pass green.

**Acknowledged, deferred (non-blocking):**
- *No pi typecheck step* — pi has no `tsconfig`/`typecheck` script today; adding one is its own small follow-up rather than CI scope. The smoke exercises the full boot path in the meantime.
- *pi `bun-version: latest` only* — intentional for an agent harness; matrixing bun adds cost without much signal here.
- *Pre-existing `smoke.mjs` `{type:"status",data:"ack"}` vs `encodeChunk({...status})` naming* — confirmed it's a wire-shape vs internal-field-name confusion (the wire field is `data`; tests are correct at 13/13). Not introduced here; left for a future cleanup pass.

Also retargeted this PR's base to `main` now that #158 is merged — `MERGEABLE/CLEAN`.

PR #160, opened 2026-06-16
Python: mirror the SYNADIA_* identity env-var convention

renerocksai:
Mirrors the TypeScript agents' `SYNADIA_*` identity env-var convention onto the two Python surfaces Rene maintains: the agent-sdk examples ladder and the deerflow channel. No other Python surface reads NATS identity.

## The convention

Identity (`owner` = 4th subject token, `name`/session = 5th token) resolves first-non-empty-wins:

```
CLI flag  >  SYNADIA_<AGENT>_<TOKEN>  >  SYNADIA_<TOKEN>  >  <legacy var>  >  config/derived default
            └ per-agent override ┘      └ fleet-wide ┘      └ kept as alias ┘
```

- `<AGENT>` = the agent's own subject token, uppercased, hyphens→underscores (e.g. `echo` → `SYNADIA_ECHO_OWNER`). It derives from the **registered** token, not the filename — `04-combined.py` registers `agent="llm"`, so its per-agent var is `SYNADIA_LLM_OWNER`.
- `<TOKEN>` = `OWNER` or `NAME` (the spec word for the 5th token, even where the code calls it `session`).
- Connection vars (`NATS_URL`, `NATS_CONTEXT`, `NATS_CREDENTIALS`) stay generic and untouched.
- The `agent` (3rd) token is **not** env-settable under this scheme. Deerflow keeps its existing `NATS_AGENT_TOKEN`/`DEERFLOW_NATS_AGENT` handling unchanged.

## Scope decision: precedence ladder ported, Python's `or`-chain idiom preserved

The TS side has a "first-present-wins, does-NOT-cascade-on-empty, loud-fail-on-invalid-token" semantic. **This PR does not port that semantic** — it ports the *precedence ladder* only. Rationale:

- The Python SDK's `AgentSubject.new` / `_sanitize` base64-encodes non-conforming tokens rather than failing loud — a different, already-shipped contract.
- Both Python surfaces already resolve identity with `or`-chaining (empty/whitespace cascades to the next source). We preserve that idiom and only **prepend the new `SYNADIA_*` sources** to the chain.

So a blank `SYNADIA_*` value cascades to the next rung rather than registering an empty owner — verified at runtime (see below).

## Change 1 — agent-sdk examples ladder

- `examples/_connect_cli.py`: `add_agent_identity_flags` gains a keyword-only `agent` param; `--owner`/`--session-name` defaults resolve through the ladder above (`agent=None` skips the per-agent rung — the reference-agent path, whose token is a runtime CLI flag). Help strings name the new vars.
- The five numbered examples pass their registered token (`echo`/`ollama`/`openrouter`/`llm`/`tools`). `_reference_agent.py` keeps `session_fallback="example"` and no per-agent var.
- `examples/README.md` env-var table + run snippets, and `CHANGELOG.md` `[Unreleased]`, updated.
- New `tests/test_connect_cli_identity.py` path-loads the helper and asserts the ladder, the `agent=None` path, flag-override, and hyphen→underscore mapping.

## Change 2 — deerflow channel

- `config.py` `_apply_env()`: prepends `SYNADIA_DEERFLOW_OWNER`/`SYNADIA_OWNER` and `SYNADIA_DEERFLOW_NAME`/`SYNADIA_NAME` to the owner/session chains; legacy `NATS_OWNER`/`DEERFLOW_NATS_OWNER`/`NATS_AGENT_NAME`/`NATS_SESSION` kept as lowest-priority aliases. `agent` token and connection vars untouched; CLI flags still win.
- `README.md` env-var table, quickstart exports, troubleshooting line updated. (Bonus readability note in the docs: `NATS_AGENT_NAME`, named "agent", actually sets the 5th *session* token — the `SYNADIA_*` names fix that confusion.)
- `tests/test_config.py`: new `test_synadia_identity_env_precedence` (full ladder + CLI override) and the default-state test now clears the new vars so a developer's shell can't leak in.

## Legacy aliases

All legacy aliases are kept everywhere as lowest-priority sources — zero behavior change for anyone already setting old vars.

## Verification

Lint/format/mypy/pytest green for both packages (deerflow: 47 passed; agent-sdk: 47 passed, synced `--all-extras` to match CI).

Beyond CI, `/verify` drove the two real CLI surfaces against a live `nats-server`:
- **deerflow `doctor`**: all five rungs + CLI override resolved to the expected `prompt_subject`; blank/whitespace `SYNADIA_*` correctly cascaded.
- **agent-sdk examples**: `01-echo` / `04-combined` / `_reference_agent` printed the expected `agents.prompt.<agent>.<owner>.<session>`. Confirmed `04-combined`'s per-agent var keys off the registered `llm` token (a filename-derived `SYNADIA_COMBINED_OWNER` was ignored), and the reference agent (`agent=None`) ignored a per-agent var while honoring `SYNADIA_OWNER` + the `example` fallback.

## Ripple-check

- TS SDK public surface: untouched (Python-only).
- Wire format / protocol behavior: unchanged — this is config resolution, not wire shape. No interop-test impact.
- READMEs (agent-sdk examples + deerflow) and agent-sdk CHANGELOG updated. Deerflow has no CHANGELOG (README is its doc surface).
- No release ladder needed — examples consume the local SDK; deerflow uses its own config module.

Reference TS template PRs: #150 (SDK examples), #153 (openclaw legacy-alias handling). The §5.3 `{"prompt":""}` decode-parity item is explicitly out of scope per the handoff.

### DIFF
--- a/agent-sdk/python/CHANGELOG.md
+++ b/agent-sdk/python/CHANGELOG.md
@@ -15,11 +15,21 @@ the 0.x line is explicitly unstable per protocol spec §11.2.
   `agent-sdk/typescript/examples/`: echo, Ollama, OpenRouter, a
   combined auto-selecting agent, and a tool-calling agent backed by a
   NATS microservice. Identity and heartbeat are flags that default to
-  the `NATS_AGENT_OWNER` / `NATS_AGENT_NAME` /
-  `NATS_AGENT_HEARTBEAT_INTERVAL` env vars (env-first, flag-overridable);
-  connection uses the shared `_connect_cli.py` resolver. The
-  `_reference_agent.py` flags now likewise default to those env vars
-  (non-breaking — explicit flags still win).
+  env vars (env-first, flag-overridable); connection uses the shared
+  `_connect_cli.py` resolver. Identity follows the same `SYNADIA_*`
+  ladder the TypeScript agents use, first non-empty wins:
+  `SYNADIA_<AGENT>_OWNER` (per-agent) > `SYNADIA_OWNER` (fleet-wide) >
+  `NATS_AGENT_OWNER` (legacy alias) > `$USER` > `anon`, and the `NAME`
+  analogue (`SYNADIA_<AGENT>_NAME` > `SYNADIA_NAME` > `NATS_AGENT_NAME`
+  > the `--session-name` fallback) for the 5th subject token.
+  `<AGENT>` is the example's registered subject token uppercased with
+  hyphens turned into underscores (e.g. `echo` → `SYNADIA_ECHO_OWNER`).
+  Heartbeat cadence stays config, not identity — it keeps its single
+  `NATS_AGENT_HEARTBEAT_INTERVAL` var. The `_reference_agent.py` flags
+  default through the fleet-wide + legacy + fallback rungs only (no
+  per-agent var — its `agent` token is a runtime CLI flag). Non-breaking
+  — explicit flags still win, and the legacy `NATS_AGENT_*` vars keep
+  working as lowest-priority aliases.
 - **`examples` extra** — `httpx`, used by the LLM/tool example scripts
   (`uv sync --extra examples`). Not part of the published SDK surface.
 
--- a/agent-sdk/python/examples/01-echo.py
+++ b/agent-sdk/python/examples/01-echo.py
@@ -6,10 +6,11 @@
 # agent-sdk/typescript/examples/01-echo.ts.
 #
 # Identity → subject agents.prompt.echo.<owner>.<session_name>. Owner and the
-# session name are overridable (--owner / --session-name, or $NATS_AGENT_OWNER /
-# $NATS_AGENT_NAME) so several people can run this against one server without
-# colliding. --heartbeat-interval / $NATS_AGENT_HEARTBEAT_INTERVAL tunes the
-# heartbeat cadence (default 30s).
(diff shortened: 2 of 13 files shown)


renerocksai:
@claude can you please review

renerocksai:
Thanks for the review — all three findings addressed in `28b079a`:

1. **Test isolation (deerflow)** — promoted the per-test `delenv` to an `autouse` fixture that clears `_IDENTITY_ENV_VARS` before every test, mirroring the agent-sdk `test_connect_cli_identity.py` pattern. `test_config_file_then_env_then_cli_precedence` (and every other test) now can't be flaked by a stray `SYNADIA_*` export in the dev's shell.
2. **Stale `doctor.py` error message** — now reads `owner is not configured; set SYNADIA_DEERFLOW_OWNER, SYNADIA_OWNER, a legacy NATS_OWNER/DEERFLOW_NATS_OWNER alias, or pass --owner`, and the README troubleshooting entry's literal was synced to match. Verified the new text renders at the CLI (`doctor` with no owner set).
3. **Stale header comments** — `01-echo.py` and `04-combined.py` now lead with the `SYNADIA_*` ladder instead of the legacy `NATS_AGENT_*` vars (04 notes its registered `agent="llm"` token, so the per-agent var is `SYNADIA_LLM_OWNER`).

Re-ran lint/format/mypy/pytest for both packages — all green.

renerocksai:
@claude can you please re-review — the three findings from your last pass are addressed in 28b079a.

PR #162, opened 2026-06-16
ci(deerflow): add PR-level CI for agents/deerflow

renerocksai:
Adds the missing PR/push CI for the DeerFlow channel.

## Why

deerflow had **no PR-level CI**. Its test suite only ran inside the `verify` job of `release-python-deerflow.yml`, which fires on `python-deerflow-v*` tags. So a change could land on `main` — as the `SYNADIA_*` identity work in #160 did — with deerflow's 47 tests never executing in CI until release time. This closes that gap so every `agents/deerflow/**` PR runs the suite.

## What

A `agents/deerflow/**`-scoped workflow whose steps are **intentionally identical** to the release `verify` job:

```
uv sync --all-extras
uv run ruff check .
uv run ruff format --check .
uv run mypy src tests
uv run pytest -v          # py3.11 / 3.12 / 3.13
```

A green PR run therefore predicts a clean publish: if this passes, the tag's pre-publish gate will too.

## Scope decisions

- **deerflow-only path filter** (`agents/deerflow/**` + this workflow file). deerflow resolves `synadia-ai-agents` / `synadia-ai-agent-service` to the local checkouts via `[tool.uv.sources]`, so a breaking SDK change *could* affect it — but those SDKs have their own CI, and deerflow's release `verify` job is the backstop. Running deerflow's full matrix on every SDK PR would be redundant.
- **No `nats-server` step** (unlike `client-sdk-python-agent-service.yml`): deerflow's suite has no broker-spawning e2e tests, so the binary isn't needed. Verified locally — the 47 tests run in ~0.2s with no skips.

## Validation

YAML parsed and structure checked locally; run-commands diffed against the release `verify` job (identical). This workflow's own first run will be on this PR (it path-matches itself), so the checks below are its live proof.

### DIFF
--- a/.github/workflows/agents-deerflow-python.yml
+++ b/.github/workflows/agents-deerflow-python.yml
@@ -1,0 +1,67 @@
+name: CI — agents/deerflow
+
+# PR/push CI for the DeerFlow channel (`synadia-ai-nats-deerflow-channel`).
+#
+# Until now deerflow had no PR-level CI: its test suite only ran inside the
+# `verify` job of `release-python-deerflow.yml`, which fires on
+# `python-deerflow-v*` tags — so a change could land on `main` (as the
+# SYNADIA_* identity work in #160 did) with deerflow's tests never executing
+# in CI until release time. This workflow closes that gap.
+#
+# The steps are intentionally identical to that release `verify` job, so a
+# green run here predicts a clean publish: if PR CI passes, the tag's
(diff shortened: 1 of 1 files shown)


renerocksai:
@claude please review — adds the missing PR-level CI for agents/deerflow so its suite runs on PRs, not only at release-tag time. (Manual ping; the bot no longer auto-starts.)

PR #164, opened 2026-06-29
Load NATS context TLS files and honor agent web UI host binding

escidmore:
## Summary

This makes two small compatibility fixes for Bun/Node clients that use NATS CLI contexts and the agent web UI:

- Load NATS context TLS file paths into standard TLS option contents.
- Add an explicit host bind option for `examples/agent-web-ui`.

### NATS context
NATS contexts created with:

```bash
nats context add --tlscert ... --tlskey ... --tlsca ...
```

store cert, key, and ca as file paths. The SDK previously passed those through as certFile, keyFile, and caFile. Some runtimes/transports do not consume those helper fields and instead expect the standard TLS options:

tls.cert
tls.key
tls.ca

This caused Bun-based clients using a valid NATS context with client mTLS to fail even though the same context worked with the NATS CLI.

This PR reads the configured TLS files during context loading and passes their contents through standard TLS options. File read failures are wrapped in NatsContextError so context resolution still fails with an SDK-specific error.

### --host acceptance
The agent web UI also documented/accepted --host, but the value was not passed to Bun.serve(). This meant commands like:
NATS_CONTEXT=prod bun run start --host 10.0.1.9

did not actually bind to the requested interface. This PR passes hostname only when explicitly configured, preserving Bun’s default binding behavior otherwise. The app-specific AGENT_WEB_UI_HOST environment variable is also supported.

## Compatibility
This PR keeps the diff intentionally narrow:

- Does not change the NATS context file format.
- Does not disable TLS hostname verification.
- Does not pass hostname to Bun.serve() unless --host / AGENT_WEB_UI_HOST is set.
- Avoids the generic HOST environment variable to prevent platform/CI/container collisions.
- Preserves the existing default server binding behavior.

The main behavior change is that missing or unreadable TLS files now fail during loadContextOptions() instead of later during connect(). Those failures are wrapped as NatsContextError.

## Verification
```bash
bun test client-sdk/typescript/test/unit/context.test.ts

33 pass
0 fail

cd client-sdk/typescript
bun run typecheck
bun run build

cd examples/agent-web-ui
bun run typecheck
bun run build
```

Live smoke checks:
```bash
❯ NATS_CONTEXT=homelab-audit-webui bun run start --host 10.0.1.9
$ bun run server/index.ts --host "10.0.1.9"
[testui] NATS client connected (context=homelab-audit-webui)
[testui] listening on http://10.0.1.9:3300 (sdk protocol 0.3)
```

### DIFF
--- a/client-sdk/typescript/src/context.ts
+++ b/client-sdk/typescript/src/context.ts
@@ -26,7 +26,7 @@
 //
 // Supported context fields: `url`, `creds` (path), `nkey` (path),
 // `user_jwt` (+ optional `user_seed` for nonce signing), `user`+`password`,
-// `token`, `inbox_prefix`, plus the TLS triple `cert`/`key`/`ca` and
+// `token`, `inbox_prefix`, plus the TLS file triple `cert`/`key`/`ca` and
 // `tls_first`.
 // Skipped: `nsc` integration.
 //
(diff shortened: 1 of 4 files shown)


M64GitHub:
@claude please review

M64GitHub:
Thanks for this fix, @escidmore — it's correct and we want it in. Since this branch predates the prettier 3.9.5 pin that landed in #166, its CI can no longer pass as-is (the new test file needs a one-spot reformat), and the repo conventions also want a `CHANGELOG.md` entry for the `loadContextOptions` behavior change.

Rather than push to your branch, we've opened #167, which cherry-picks your commit unchanged (authorship preserved) and adds only the formatting pass, the CHANGELOG entry, and a `--host` row in the web-ui README. Verified on current main: full client-sdk suite (unit under Node + Bun, integration against a real nats-server), agent-sdk suite, and web-ui typecheck/build.

Once #167 merges, this PR can be closed. Thanks again!

PR #166, opened 2026-07-13
feat(examples): 02-ollama rides the OpenAI-compatible endpoint — OLLAMA_URL accepts proxies

M64GitHub:
## Summary

Switches the `02-ollama` example's `ollamaTokens` from Ollama's native `/api/generate` (NDJSON) to its OpenAI-compatible `/v1/chat/completions` (SSE — same parsing as `03-openrouter.ts`).

Against a bare local Ollama the behavior is unchanged. The gain is that `OLLAMA_URL` now accepts anything speaking the OpenAI wire shape — e.g. an OpenAI-style metering/audit proxy (a synadia-model-proxy per-agent identity URL) in front of the local instance, which can then meter the streamed call token-exactly.

- `agent-sdk/typescript/examples/02-ollama.ts` — request body switched to `messages`/chat shape, streaming parser switched from NDJSON to SSE (`data: {json}` / `data: [DONE]`), header comments updated.
- `agent-sdk/typescript/examples/README.md` — `OLLAMA_URL` description now reflects that a proxy works too.

## Verification

Verified live: `bun examples/02-ollama.ts` through synadia-model-proxy to a local Ollama (llama3.1:8b) — streamed reply over NATS, exact usage record. prettier + `tsc --noEmit` + eslint clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #167, opened 2026-07-13
Load NATS context TLS files and honor agent web UI host binding (supersedes #164)

M64GitHub:
## Summary

Supersedes #164 — all credit to @escidmore, whose commit is cherry-picked here unchanged (authorship preserved). This PR adds only what was needed to land it on current `main`: a prettier-3.9.5 formatting pass on the new test (the branch predates the pin from #166), the `CHANGELOG.md` entry required for public-SDK-surface changes, and the web-ui README row for the new `--host` flag.

### The fix itself (from #164)

- `loadContextOptions` now reads the TLS triple's file contents (`cert`/`key`/`ca` context fields) into the standard `tls.cert`/`tls.key`/`tls.ca` options instead of passing paths via the Node-only `certFile`/`keyFile`/`caFile` helper fields. `@nats-io/transport-node`'s `loadClientCerts()` supports the contents form first-class, and it's portable to runtimes that never consumed the helper fields (e.g. Bun) — a context created with `nats context add --tlscert/--tlskey/--tlsca` now works everywhere.
- The agent web UI's documented `--host` flag is now actually passed to `Bun.serve()` (plus `AGENT_WEB_UI_HOST`); binding behavior is unchanged unless explicitly set.

### Behavior notes (also in the CHANGELOG)

- Missing/unreadable TLS files now throw `NatsContextError` (with `cause`) at `loadContextOptions()` instead of failing inside `connect()` — consistent with the existing eager reads of `creds`/`nkey` files, and it turns the "stuck reconnecting on a bad cert path" failure mode (0.5.2 behavior notes) into a fast, actionable error.
- TLS material is snapshotted at load time; long-lived processes no longer pick up rotated cert files on reconnect the way per-attempt `certFile` reads did. Re-run `loadContextOptions()` to refresh.
- Repo sweep found no consumers introspecting `tls.certFile`/`keyFile`/`caFile` from `loadContextOptions` output; `agents/pi` and `agents/claude-code` have their own context loaders and are unaffected (they still carry the Bun mTLS limitation — follow-up candidate). The Python SDK rejects TLS-triple contexts as unsupported, so no cross-SDK parity impact.

## Verification (on this branch, current main base)

- `client-sdk/typescript`: `format:check`, `typecheck`, `lint` clean; **254/254 unit** (Node + Bun) and **41 integration** tests pass against a real nats-server.
- `agent-sdk/typescript`: `format:check`, `typecheck`, `lint`, 20/20 unit tests.
- `examples/agent-web-ui`: `typecheck` + `build` clean against the freshly built `file:`-linked SDK (no CI covers this example — verified locally).
- Original live mTLS smoke through a real context: see #164's verification section.

### DIFF
--- a/client-sdk/typescript/CHANGELOG.md
+++ b/client-sdk/typescript/CHANGELOG.md
@@ -34,6 +34,29 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
   missing / non-string / empty `prompt` and invalid attachments. The pi
   spec-compliance smoke (`agents/pi/test/smoke.mjs`) caught both gaps.
 
+- **`loadContextOptions` now loads the TLS triple's file contents**
+  (the `cert` / `key` / `ca` context fields) into the standard
+  `tls.cert` / `tls.key` / `tls.ca` connection options, instead of
+  passing paths through the Node-only `certFile` / `keyFile` / `caFile`
+  helper fields. Fixes client mTLS on runtimes whose transports don't
+  expand the helper fields (e.g. Bun) — a context created with
+  `nats context add --tlscert/--tlskey/--tlsca` now works everywhere
+  the standard options do. See
+  [#164](https://github.com/synadia-ai/synadia-agents/pull/164).
+
+  Behavior notes:
+
+  - Missing or unreadable TLS files now throw a `NatsContextError`
+    (with `cause`) from `loadContextOptions()` instead of failing
+    later inside `connect()`. This matches how `creds` / `nkey` files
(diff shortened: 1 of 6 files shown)


M64GitHub:
@claude please review

M64GitHub:
Both review suggestions addressed in dfaac23: added the `tls_first: true`-only case (asserts `opts.tls = { handshakeFirst: true }` with no file reads) and the partial-triple case (`ca` only → `opts.tls = { ca: ... }`), exercising the independent guards. 35/35 context tests pass; CI green.

PR #168, opened 2026-07-13
Context-loader follow-ups: consistent errors, agent TLS ports, examples CI

M64GitHub:
## Summary

Sweeps the follow-up list from the #164/#167 investigation. Independent pieces, one commit each:

### 1. `client-sdk`: consistent context-load failures (`a2be41b`)
- `creds`/`nkey` read failures now throw `NatsContextError` with the filesystem error as `cause`, matching the TLS triple from #167 — previously raw `ENOENT` escaped `loadContextOptions()`. Callers matching `err.code === "ENOENT"` should inspect `err.cause` (CHANGELOG'd under `[Unreleased] > Changed`).
- Documented the path-resolution rules in the module header (leading `~/` expands; other relative paths resolve against process CWD, as the `nats` CLI does — the footgun the #167 review flagged). Behavior unchanged, now written down.

### 2. `agents/pi` + `agents/claude-code`: TLS contents port (`6db6355`, `378db05`)
The two hand-rolled context loaders still passed `certFile`/`keyFile`/`caFile` — Node-only helper fields Bun's transport never consumed — so mTLS contexts failed there even after #167 fixed the SDK. Both now read PEM contents into standard `tls.cert`/`key`/`ca` (forwarding the original read error as `cause`, per review), and both fix a shared latent gap: `tls_first` **alone** previously never produced `handshakeFirst` (the guard ignored it without cert/key/ca).

PI's loader is now exported and covered by a new 5-case `test/context.test.ts`, wired into the CI unit-test step. claude-code has no test infra; its port is line-for-line the SDK's typechecked pattern and passes a parse check.

Not in scope (tracked in the follow-up issue): migrating these loaders to the SDK's `loadContextOptions`, their `user_jwt`-without-seed gap, and their missing `~/` expansion. The other five agents already use the SDK loader and get the fix via the next SDK release + bump.

### 3. CI for `examples/*` (`6d0a844`, `340867a`, `901d931`)
`examples/*` had zero CI. New `examples-typescript.yml` builds both `file:`-linked SDKs and typechecks each of the seven examples, on changes to `examples/**` or either SDK. This PR itself proves the gap: **`pi-headless` was silently broken on main** — it imports `@earendil-works/pi-ai` without declaring it (worked only via transitive hoisting that clean installs no longer provide). Declared aligned with `pi-coding-agent`'s own range (`^0.79.1`). The `open-agent-vercel` job additionally installs `agents/open-agent`'s deps, since that example's tsconfig path-aliases into the agent's vendored sources.

**The workflow installs with npm, not bun** — and this PR's own first CI run is the evidence: `bun install` in the agent-sdk step stalled ~7 minutes and died with `Illegal instruction (core dumped)` (exit 132) in 4 of 7 identical concurrent jobs ([the run](https://github.com/synadia-ai/synadia-agents/actions/runs/29242174408)), while the same step passed elsewhere. Combined with the local macOS behavior below, bun 1.3.14's resolver is **nondeterministic on the client-sdk ⇄ agent-sdk `file:` cycle — cross-platform**: it can succeed, hang forever, or crash. npm symlinks `file:` deps and resolves the same graph in seconds.

## Verification

- client-sdk: format/typecheck/lint clean, **258/258 unit**, **41 integration** against a real nats-server; agent-sdk suite clean (fires on client-sdk src changes).
- agents/pi: **14/14 unit** (owner + new context tests), **13/13 spec smoke**.
- All **seven examples typecheck** (pi-headless after the dep fix); the workflow validates itself on this PR.

## Heads-up for anyone installing locally

The bun bug above bites dev machines too: on macOS, `bun install` **hangs forever** (~100% CPU, threads parked in `kevent64`) in any package needing fresh resolution of the SDK `file:` cycle — which since the #166 prettier pin staled every dependent lockfile means effectively all of them (agents/pi, examples/*). Latest bun (1.3.14) is affected; there is no newer release. Working fallback: `rm -rf node_modules && npm install`, keep bun as the runtime; don't commit the resulting `package-lock.json`, and note npm runs postinstall scripts bun blocks (the Ax packages drop `.claude/skills/` dirs into `examples/dspy*` — delete them).

### DIFF
--- a/.github/workflows/agents-typescript.yml
+++ b/.github/workflows/agents-typescript.yml
@@ -54,9 +54,9 @@ jobs:
       - name: Install pi
         working-directory: agents/pi
         run: bun install
-      - name: Unit tests (owner-token precedence)
+      - name: Unit tests (owner-token precedence + context TLS)
         working-directory: agents/pi
-        run: bun test test/owner.test.ts
+        run: bun test test/owner.test.ts test/context.test.ts
       - name: Spec-compliance smoke (spawns its own nats-server)
         working-directory: agents/pi
         run: bun test/smoke.mjs
--- a/.github/workflows/examples-typescript.yml
+++ b/.github/workflows/examples-typescript.yml
@@ -5,6 +5,12 @@ name: CI — examples (TypeScript)
 # here instead of after a release. Typecheck-only by design: the
 # examples talk to live NATS servers / LLM backends, so their runtime
 # behavior is exercised manually (see each example's README).
+#
+# Installs use npm, not bun: bun 1.3.14's resolver is nondeterministic
+# on the client-sdk ⇄ agent-sdk `file:` cycle — it can hang forever or
+# die with SIGILL ("Illegal instruction") after ~7 minutes, on macOS
+# and Linux alike. npm symlinks `file:` deps and is unaffected. Revisit
+# once a bun release fixes cyclic file: resolution.
 
 on:
   push:
@@ -42,17 +48,17 @@ jobs:
           - pi-headless
     steps:
       - uses: actions/checkout@v4
-      - uses: oven-sh/setup-bun@v2
+      - uses: actions/setup-node@v4
         with:
-          bun-version: latest
+          node-version: 22
       # The examples import the SDKs via `file:` links, which resolve to
       # each package's built `dist/` — build both before installing.
       - name: Build client-sdk
         working-directory: client-sdk/typescript
-        run: bun install && bun run build
+        run: npm install --no-audit --no-fund && npm run build
       - name: Build agent-sdk
         working-directory: agent-sdk/typescript
-        run: bun install && bun run build
+        run: npm install --no-audit --no-fund && npm run build
       - name: Typecheck
         working-directory: examples/${{ matrix.example }}
-        run: bun install && bun run typecheck
+        run: npm install --no-audit --no-fund && npm run typecheck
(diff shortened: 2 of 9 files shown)


M64GitHub:
@claude pelase review

M64GitHub:
Both pre-merge items addressed:

1. **`{ cause }` in the agents' `readTlsFile`** — fixed in 378db05: both now throw `new Error(\`failed to read TLS ${field} file ${path}\`, { cause: err })`, matching the SDK's structured form. PI tests re-run green (14/14); claude-code parse check clean.

2. **Workflow comment vs PR description** — good catch, but the resolution is inverted: the **comment is correct and the PR description was stale**. The description's "Linux CI is unaffected" line was written before this PR's first CI run, in which `bun install` died with `Illegal instruction (core dumped)` / exit 132 on ubuntu-24.04 in 4 of 7 concurrent jobs (https://github.com/synadia-ai/synadia-agents/actions/runs/29242174408) — same cyclic-`file:` resolution bug that hangs forever on macOS, hence "nondeterministic … on macOS and Linux alike." The PR description is now updated to match the evidence; `agents-typescript.yml`'s bun installs pass today the same way they always have — lockfile fast paths and luck — and remain flake-exposed until bun fixes cyclic `file:` resolution.

The three follow-up notes (agents' `user_jwt`-without-seed gap, missing `~/` expansion on context file paths, raw creds/nkey errors in the agents' loaders) are acknowledged as real parity gaps vs the SDK loader — they're candidates for either in-place fixes or the larger "migrate hand-rolled loaders to `loadContextOptions`" discussion per CLAUDE.md.

PR #182, opened 2026-08-29
ts(identity): PR-T2 — verifySender, resolveSender/resolve(), operatorAttested, --signed probe, T5 operator fixture

renerocksai:
PR-T2 of the sender-identity block (plan §7 step 4 in `synadia-agent-fabric-docs/docs/plans/agent-identity-sdk-implementation-plan.md`; spec `docs/agent-protocol-sender-identity.md`). Base: `main` at `928ed2f`. Ships with PR-T1 as `0.6.0`; CHANGELOGs stay under `[Unreleased]`.

## What lands

**Client `@synadia-ai/agents`**
- `verifySender(msg, "live" | "stored", opts?)` — the spec-shaped `VerifySender` over a structural `{ subject, data, headers? }` (`Msg`, `ServiceMsg`, `JsMsg`); `undefined` for no header / unknown `v`. The JetStream test now consumes it.
- Reverse lookup (`ResolveSender`): `agents.resolveSender(id)` + `new Agents({ resolveTtlMs })` (default 10 s), `SenderResolver` (`resolve` / `invalidate` / `ttlMs`; one in-flight enumeration shared by concurrent callers, a settled one always dropped), the uncached `resolveSender(nc, id)`. Same shape as Python's `SenderResolver` (PR-P1's §12 finding). `discover()` and the resolver share a new `enumerateAgentInfos`.
- Operator-attested mode in the codec (spec Appendix A): `VerifySenderOptions.operatorAttested` (+ `headers`), `src/identity/request-info.ts` (`NATS_REQUEST_INFO_HEADER`, `parseRequestInfo`, `readRequestInfo`). `Nats-Request-Info` is read **only** under that mode. `VerifiedSender.accountAttested` is now set when the stamp's `acc` — or the `accountTokenPosition` cross-check — agrees with the signed account.
- `examples/_run-client-probe.ts`: `--signed`, `NATS_NKEY_SEED_FILE` / `NATS_CREDS`, a leading `{"type":"identity","id":…}` line under `--signed` (not counted in `done.chunks`), `name` / `code` on the stderr error line. Without `--signed` the probe sends **no** header at all (not even a claim), so PR-P2's reverse leg gets exactly "verified" vs "absent". The existing Python reverse-interop test passes unchanged.

**Host `@synadia-ai/agent-service`**
- `AgentServiceOptions` / `ReferenceAgentOptions`: `resolveTtlMs`, `operatorAttested` (validated). `PromptResponse.sender.resolve()` is bound to a `SenderResolver` on the host's connection. `SenderGate.classify` runs `verifySender(msg, "live", …)`; `SenderGateOptions.operatorAttested` / `.resolver`.

**Fixture (optional T5)** — `test-fixtures/identity/operator/`: operator JWT, `resolver: MEMORY` + preloaded SYS / ACME / APP account JWTs, creds for alice / carol (ACME) and bob (APP, `share: true` imports of `agents.>` and `$SRV.>`), generated by `@nats-io/jwt` in its own `package.json`. Port-less; lints with `nats-server -t`; the four CI lint steps now include `operator/*.conf`. Test-only keys, committed (same policy as `keys.json`).

## Decisions where the plan was silent (recorded in the plan's §12 on this PR)

- The cross-check runs **before** the nonce lookup and the signature (cheap first, same `401`); it applies to **signed headers only** — claims are never cross-checked (the spec's sentence is about "a verified signature and a declared-trustworthy header").
- A stamp that is present but not parseable as the server's (`{bad`, non-object, non-string `acc`/`user`, or two `Nats-Request-Info` values) is a `401` under the mode — the spec defines only disagreement; under a closed-endpoint promise a stamp the server did not write is itself evidence of forgery.
- `accountAttested = operatorAttested && (accountTokenPosition configured || stamp.acc present && equal)` — the plan's §2.3 formula; a present `user` must also agree but adds nothing to attestation (it is already verified).
- `readRequestInfo` returns `undefined` (absent) / `null` (malformed) / the stamp — no ninth error class.
- `SenderResolver` re-parses an `AgentId | string` argument (`InvalidAgentIdError` on garbage), like Python's `AgentId.parse`.
- `resolveTtlMs` is validated by `normalizeResolveTtlMs` (finite, ≥ 0) and throws `IdentityError` from both `Agents` and the host classes.

## Verification (local: bun 1.3.14, nats-server 2.12.7)

- client: typecheck / lint / prettier clean; unit 403 passed (incl. the new `verify-sender.test.ts`, 16); integration 97 passed / 2 pre-existing websocket skips — `identity-accounts` (operator-attested T2/T3/T4 rows + `resolveSender` end to end), `identity-jetstream` on `verifySender(m, "stored")`, the new `identity-t5-operator` (5).
- host: clean; unit 29, integration 39 passed (resolve() binding, cross-account attested rows, option validation).
- `agent-sdk/python` `test_interop_reverse_e2e.py`: 2 passed against the modified probe.
- Manual: reference agent with `NATS_NKEY_SEED_FILE` + `REFERENCE_AGENT_MIN_SENDER_TRUST=signed`; probe `--signed` → identity line + `… (verified user, claimed account)` echo, exit 0; unsigned → `SenderSignatureRequiredError`, exit 1; `--signed` without a seed → exit 64.


renerocksai:
@claude the findings are addressed in `6406b64` (CI green, 33/33 on that commit). Could you confirm, one line per finding:

1. **`#capWarned` one-shot** — `NonceCache.sweep()` now re-arms the warning once normal expiry has drained the set to half the cap (hysteresis on purpose: an eviction round itself leaves the set just under the cap, so a plain "below the cap" reset would log on every round); new unit test `the cap warning is re-armed once expiry drains the set to half the cap — not by eviction rounds alone`.
2. **`classify` refactor** — ✓ as you read it; no change.
3. **Nonce CAS comment** — the comment in `admitPrompt` now states why it is atomic: no `await` between `classify()` resolving and `record()`, the event loop cannot interleave another request's `record()`, and the earlier `nonceSeen` lookup inside `classify()` (across awaits) is only the cheap early exit.
4. **`operatorAttested` is a deployment promise** — agreed and unchanged; the option's JSDoc, the host README ("On an open endpoint … leave it off: a peer can write that header, and the mode would attest a forgery") and both T2 tests (a same-account forger is refused only because the *account* disagrees — the T2 row that self-stamps `acc: ACME` demonstrates the exposure) carry it.
5. **`SenderResolver.#rebuild` rejection semantics** — ✓ as you read it; no change.
6. **No size cap in `parseRequestInfo`** — deliberately none, now documented in its JSDoc: behind a `share: true` import the server's stamp embeds the caller's whole user JWT, which grows with its permission lists (the T5 fixture's stamp is ~1 KB with an empty permission set), so a cap would refuse legitimate stamps; the value is parsed only under operator-attested mode, where it is the server's, and the broker's `max_payload` bounds it.
7. **Committed test keys under `test-fixtures/identity/operator/`** — same policy as `keys.json` (PR-0). The repo has no secret scanner configured (`secret_scanning: disabled`, no gitleaks/trufflehog workflow), so there is nothing to exclude today; noted in the plan's implementation log for whoever enables one.

PR #183, opened 2026-08-29
py(identity): PR-P2 — host identity in synadia-ai-agent-service 0.5.0 (floor synadia-ai-agents>=0.8)

renerocksai:
## Summary

PR-P2 of the sender-identity plan (`docs/plans/agent-identity-sdk-implementation-plan.md` §4.2 / §7 step 5 in `synadia-agent-fabric-docs`): the **receiver side** of the extension in `synadia-ai-agent-service` **0.5.0**, floor **`synadia-ai-agents>=0.8`** (both bumped here; development resolves the sibling client via `[tool.uv.sources]`, no PyPI release needed). Protocol stays `0.3` — support is advertised by feature detection (`min_sender_trust`).

### Host (`agent-sdk/python`)

- **`agent_service/identity.py`** (new): `NonceCache` (keyed `(user, nonce)`, expiry at `ts + window`, second-bucketed sweeps, hard cap with the re-armed warning, synchronous check-and-set `record()`), `SenderGate` (`classify` / `admit_prompt` / `classify_status`, the `min_sender_trust` gate, the `accept_sender` hook, the 400/401/403/500 mapping, the resolver binding), `ServiceIdentity(signer=…)`, `AcceptSenderHook`, `SenderAdmission` / `SenderRejection`.
- **`AgentService`**: classification after the envelope checks and **before the ack** (a refusal = §9 error frame + §9.3 terminator, no ack, no handler call); `PromptStream.sender`; `status` classified before the reply, never rejected; `start()` awaits `self_id` once (mismatch → `IdentityMismatchError`; no identity → logged, keys omitted), registers `user_nkey` / `account` (+ `id_sig` with a signer), **always** `min_sender_trust` on the prompt endpoint, and **flushes** before returning (PR-T1 finding). New ctor params: `identity`, `min_sender_trust`, `replay_window_s`, `account_token_position`, `accept_sender`, `resolve_ttl_s`, `operator_attested`; properties `identity`, `min_sender_trust`, `operator_attested`, `instance_id`.
- Examples: `_connect_cli.py` `--nkey` / `--creds` + `signer_from_cli`; the reference agent adds `--min-sender-trust`, prints `identity: <id> (min_sender_trust=…)` on its own line after the ready marker, and appends ` sender: <id> (<trust>)` to the echo only when a sender was classified (the TS runner's shape); the ladder examples pass the `ServiceIdentity`.
- Docs: CHANGELOG `[0.5.0]`, README "Sender identity", an agent-side `docs/protocol-mapping.md` identity table, `examples/README.md`, and the `CLAUDE.md` corrections the plan lists (`_PROTOCOL_VERSION` lives *here*, `nkey` contexts are supported, floor `>=0.8`, the package *is* on PyPI).

### Client codec (`client-sdk/python`, unreleased 0.8.0 — `python-v0.8.0` is not tagged yet)

The operator-attested cross-check PR-T2's §12 entry asked P2 to add: `verify_sender_header(…, operator_attested=, headers=)`, `verify_sender(msg, mode, operator_attested=)`, new `identity/request_info.py` (`NATS_REQUEST_INFO_HEADER`, `RequestInfoStamp`, `parse_request_info` — deliberately no size cap — and `read_request_info` → `(present, stamp)`). Rules byte for byte as TS: signed headers only, after `sub` acceptance and before the nonce lookup, a stamp the server would not write → `401`, `account_attested = position configured or (acc present and equal)`.

### Tests (local: host 85 passed, client 373 passed; ruff / format / mypy clean, cache-free)

- `test_nonce_cache.py` (the TS `nonce-cache.test.ts` rows + the re-record-after-expiry row)
- `test_identity_classify_e2e.py` — T0/T1 through `AgentService`: verified / claimed / none, `signed` rules, replay / transplant / tamper / stale-`ts`-not-recorded / malformed / unknown `v` / lowercase name, `accept_sender` 403/401/500 with the hook running before the ack, `status` classify-only with its nonce in the shared set, option validation, `resolve()` binding, "start() registers before returning"
- `test_identity_accounts_e2e.py` — carol hosts with `operator_attested=True` on `accounts.conf`: bob (share) → `(verified)`, alice (same account) → not attested + `resolve()` → her own registration, dave (no share, `acc` only) → attested, erin through the renaming import, forged / malformed `Nats-Request-Info` → 401 with the field in the log only; a second host without the mode ignores the stamp
- `test_registration_identity_e2e.py` — `$SRV.INFO` evidence: keys, `id_sig` verifies against the prompt subject only, `min_sender_trust` on prompt not status, mismatch at `start()`, T0
- reverse interop: `--signed` probe + `NATS_NKEY_SEED_FILE` → identity line `$G.<alice>`, echo `… (verified user, claimed account)`; unsigned probe vs `signed` → exit 1 `SenderSignatureRequiredError` before publishing; unsigned probe vs `any` → `sender is None`
- the reference agent under `--nkey --min-sender-trust signed`; `test_echo_e2e` metadata set → superset; client `test_request_info.py` + the operator-attested `verify` rows

### Where the plan was silent — decided (to be recorded in the plan's §12 once CI and the bot are green)

- `identity` takes a host-specific `ServiceIdentity(signer=…)` (the plan's `{ signer? }` shape), not the caller's `Identity`; passing the caller type raises `TypeError`.
- No `logger` option: the Python host keeps stdlib `logging` (`synadia_ai.agent_service.identity` / `.service`), as §4.2's parameter list has no logger.
- Classification is synchronous in Python (the verify path never awaits), so the nonce CAS shape is trivially "no `await` between the last check and `record()`"; `admit_prompt` is async only for the hook.
- `NonceCache.record()` discards an expired-but-unswept key from its old bucket before re-recording it, so the later sweep of that bucket cannot drop the fresh entry — the TS `NonceCache` has that edge (same-second re-record of an expired nonce), flagged for a later TS PR.
- §4.2 says "T2–T4 via `request_signed`": `request_signed` is single-reply, so prompts go through `Agent.prompt()` (discovery over the exported `$SRV.>`); `request_signed` is exercised against the `status` endpoint.
- The operator-attested codec change lands under the client's unreleased 0.8.0 (no 0.8.1), keeping the `>=0.8` floor exact.


renerocksai:
@claude the review findings are addressed in `1b9bc6d` (CI green on that commit: 6/6 Python jobs). Please confirm each is resolved — this is a confirmation pass, not a new review round:

1. `_admit_prompt` `respond_error` not wrapped (service.py:617 / :623) — both calls (the 500 on a classification exception, and the 400/401/403/500 refusal frame) are now inside `contextlib.suppress(Exception)`; the outer `finally` still emits the terminator.
2. `classify_status()` dead `record()` branch (identity.py:186) — documented as unreachable in a single asyncio loop (the synchronous `classify()` lookup catches replays), kept as a guard for a shared cache across threads / a future `await`.
3. `verify_sender_header` `account_attested` under an absent stamp (sender_header.py:602) — docstring now states that `operator_attested=True` alone is not sufficient; a present `acc` (or a configured position) is required, and a stampless verified header on a closed endpoint reads as traffic that did not cross the import.
4. `parse_request_info` parse-before-reject order (request_info.py:39) — noted in the docstring, with why no cap is added (`max_payload` bounds it).

PR #187, opened 2026-08-31
fix(ci): close rollout review findings

renerocksai:
Closes both follow-up findings from the completed review of #186.

- replaces the 300-file-capped push compare API with a local commit diff, fetching the exact pre-push commit when needed
- documents that open-agent integration tests own their nats-server lifecycle
- records the mandatory reviewer/follow-up loop in the persistent rollout roadmap

Validation:
- `shellcheck .github/scripts/rollout-ci-summary.sh`
- `actionlint .github/workflows/*.yml`
- `git diff --check`
- evaluator against a real docs-only push range
- evaluator against integration-only PR #171

Follow-up to inline comments on #186.

### DIFF
--- a/.github/scripts/rollout-ci-summary.sh
+++ b/.github/scripts/rollout-ci-summary.sh
@@ -28,10 +28,14 @@ else
     # rollout suite is conservative and gives the branch its first baseline.
     changed_paths=("__all_rollout_components__")
   else
-    paths_output="$(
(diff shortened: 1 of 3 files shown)


renerocksai:
@claude please review my fixes for both findings from #186. Please report any remaining issue; I will wait and address every finding before merge.

PR #188, opened 2026-08-31
fix(sdk): bind optional sender identity to live connections

renerocksai:
## Summary

- make caller and AgentService identity opt-in in both SDK languages
- bind signed identity to the live NATS user and account, with no silent downgrade
- harden identity caching, reconnect timing, malformed JWT handling, replay redaction, and hook logging
- document identity-free, unsigned, and signed use in product-neutral SDK docs
- update the rollout roadmap with the completed identity-correctness gates

## Identity contract

- omitted caller identity: no lookup, background work, or `Agent-Sender` header
- explicit unsigned identity: best-effort live lookup and claim
- configured signer: live user/account binding is mandatory for every signed path
- omitted host identity: no own lookup or registration identity metadata; inbound classification remains active
- explicit host signer: startup fails if live binding cannot be established

## Validation

- TypeScript caller: typecheck, lint, format, 509 passed / 2 skipped
- TypeScript AgentService: typecheck, lint, format, 70 passed
- Python caller: Ruff, strict mypy, 376 passed
- Python AgentService: Ruff, strict mypy, 86 passed
- `git diff --check` clean
- no private launch terminology in either public SDK tree

Targets the persistent rollout integration branch; it is not a release or main-branch cutover PR.


renerocksai:

PR #189, opened 2026-08-31
feat(sdk): resolve connection identity from one credential snapshot

renerocksai:
## Summary

- add shared TypeScript and Python connection-bundle helpers
- derive optional sender signers from the exact connection-auth snapshot
- preserve identity-off behavior for anonymous, token, password, and seedless JWT auth
- capture reconnect auth, redact representations and URL errors, and provide explicit cleanup
- keep WebSocket paths and queries intact and reject ambiguous credential sources
- document the product-neutral Agent client SDK usage and make helpers mandatory for integrations in the rollout roadmap

## Validation

- TypeScript caller SDK: typecheck, lint, format, 525 passed / 2 skipped, build
- Python caller SDK: ruff, strict mypy, 393 passed, wheel build
- TypeScript AgentService SDK: 70 passed and build
- Python AgentService SDK: 86 passed
- nats-py 2.7.0: no-nonce callback shape and real reconnect after credential-file rotation
- clean install/import from the packed npm artifact and Python wheel

## Contract

Identity is off by default. Signed mode only succeeds when the selected connection credential source contains the connection user seed. There is no second identity-only credentials path. Integrations must consume these helpers instead of copying context, creds, seed, authenticator, or signer resolution.

### DIFF
--- a/client-sdk/python/CHANGELOG.md
+++ b/client-sdk/python/CHANGELOG.md
@@ -8,6 +8,20 @@ the 0.x line is explicitly unstable per protocol spec §11.2.
 
 ## [Unreleased]
 
+### Added
+
+- `resolve_nats_connection_bundle(...)` snapshots a selected `nats` CLI
+  context or direct URL plus `.creds` / nkey connection source exactly once.
+  `identity="off"` is the default and exposes no signer;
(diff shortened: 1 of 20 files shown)


renerocksai:

PR #198, opened 2026-09-17
fix(agent-service): required registration keys win over extraMetadata

M64GitHub:
## Summary

This is the follow-up noted in #197. `AgentService` (`agent-sdk/typescript/src/service.ts`) and `ReferenceAgent` (`src/testing/reference-agent.ts`) built the registration metadata as `{agent, owner, protocol_version, ...extraMetadata}`. That let an `extraMetadata` entry replace `agent`, `owner` or `protocol_version`, so a harness could advertise an agent or owner other than the subject it serves. `session` (when set) and the identity keys (`user_nkey`, `account`, `id_sig`) already won over `extraMetadata`.

- The spread now comes first, so the required keys win too. This matches the precedence #197 gives Python's `extra_metadata`.
- The `extraMetadata` JSDoc on both option types documents the precedence.
- The `[Unreleased] / Changed` entry is in `agent-sdk/typescript/CHANGELOG.md`.
- No in-repo harness or example sets `agent` / `owner` / `protocol_version` / `session` through `extraMetadata` (acp, codex, opencode, eve, flue, openclaw, pi, both headless controllers checked). The change only affects callers that tried to override them.
- **One edge case left unchanged:** in TS, `session` is optional. When `session` isn't passed, an `extraMetadata.session` entry is still registered, as before. This PR only reorders the spread and doesn't strip that key.

## Claude Code runtime bundle

`agents/claude-code/runtime/server.js` bundles `AgentService`, so it is rebuilt here. The only bundle change is the reordered spread: 80 bytes, same file size.

How I rebuilt it: CI's `claude-code` job steps (pack the caller SDK, pack the host SDK, `npm install`, `bun run build`, `verify:bundle`), outside the repo, with bun 1.3.14. The transitive externals were held at the versions in `agents/claude-code/bun.lock`. With those pins, the **unchanged base (8cba7b5) rebuilds byte-for-byte** into the committed bundle.

⚠️ **The `claude-code` CI job will likely fail, for a reason this PR didn't introduce.** That job runs `npm install --no-package-lock`, which resolves the newest external versions. Since the last green run on 2026-09-01, several transitive packages of `@modelcontextprotocol/sdk` have been published:

| Package | `bun.lock` | Latest |
| --- | --- | --- |
| hono | 4.13.5 | 4.13.8 |
| jose | 6.2.10 | 6.2.12 |
| fast-uri | 3.1.6 | 3.1.8 |
| ip-address | 10.5.0 | 10.7.2 |
| proxy-addr | 2.0.7 | 2.0.8 |
| express-rate-limit | 8.6.2 | 8.7.0 |
| qs | 6.15.3 | 6.16.0 |
| range-parser | 1.2.1 | 1.3.0 |

With those newer versions, rebuilding the **unchanged base** no longer matches the committed bundle, so `git diff --exit-code -- runtime/server.js` fails on the base as well. I did not fold these into this bundle: they are unreviewed external updates, some only two days old, which the rollout's dependency freeze and cooldown don't allow. The CI job needs a locked install (or a reviewed dependency refresh) as a separate change.

## Tests

- `agent-sdk/typescript/test/integration/agent-service.test.ts`: forged `agent` / `owner` / `session` / `protocol_version` in `extraMetadata` are overwritten, `role: "controller"` is kept, and discovery by the forged agent returns nothing.
- `client-sdk/typescript/test/integration/discovery.test.ts`: the same check for `ReferenceAgent`, as discovered by the caller.
- Mutation check: with the old spread order restored, both new tests fail.

## Verification (local, macOS)

- `bun run check` in `agent-sdk/typescript` (typecheck, lint, format, tests): 76 passed.
- `bun run check` in `client-sdk/typescript`: 529 passed, 2 skipped (`websocket-demo.test.ts`, which is skipped by an environment gate).
- Claude Code bundle rebuild as described above; `verify:bundle` passes.
- `python3 devtools/release/release.py freeze check`: matches the baseline.
- `python3 devtools/release/release.py validate-source`: 25 classified manifests validated.

## Rollout notes

- This is an SDK feature PR, with no version changes. Release preparation already has to pick a new unique version for the Claude Code marketplace pair because its contents changed; this PR is one more reason.
- The wire protocol stays `0.3`.


M64GitHub:
@claude please review this PR. The automatic review on 4cdef45 stopped at the turn limit (35) before posting findings.

To keep it within budget: `agents/claude-code/runtime/server.js` is a ~734 KB single-line minified bundle, so its diff is unreadable. Its only change is the reordered `extraMetadata` spread inside the bundled `AgentService.start()` (80 bytes, same file size). Search it for `extraMetadata,agent:` rather than paging through `gh pr diff`. The reviewable changes are `agent-sdk/typescript/src/service.ts`, `agent-sdk/typescript/src/testing/reference-agent.ts`, `agent-sdk/typescript/test/integration/agent-service.test.ts`, `client-sdk/typescript/test/integration/discovery.test.ts`, and `agent-sdk/typescript/CHANGELOG.md`.

PR #196, opened 2026-09-16
Add tracing capabilities to PI, Claude Code, OpenClaw

cozis:
This change adds tracing capabilities to PI, Claude Code, OpenClaw similar to those implemented for the SDK in PR #194 (this builds on top of it)

PR #194, opened 2026-09-02
Add opt-in observability to client and agent SDKs

cozis:
This PR adds tracing hooks to the agent and client SDKs.

Users can now reconstruct all client-to-agent and agent-to-agent activity by enabling traces in clients and agents:

```py
# Agent
service = AgentService(
    ...
    trace=TraceOptions(), # <---
)

# Client
agents = Agents(
    ...
    identity=Identity(signer=...), # Tracing requires identity
    trace=TraceOptions() # <---
)
```

Add tracing headers to HTTP requests for model providers (`X-Synadia-Thread-ID` and `X-Synadia-Root-ID`) and pass the tool call ID to client instances when using them as tools:

```py
async def handler(envelope: Envelope, stream: PromptStream) -> None:
    response = await llm.chat.completions.create(
        ...,
        extra_headers=stream.trace_headers(),  # <--- X-Synadia-Thread-ID, X-Synadia-Root-ID
    )

    async for chunk in sub_agent.prompt(..., tool_call_id=call.id): # <---
        ...
```

When tracing is enabled, a "thread_id" is associated to each prompt. A thread is intended as all activity performed by an agent in response to a prompt, and any sub-prompt start sub-threads. The HTTP headers mark requests to the model provider with their thread ID while clients upon sending a prompt emit an edge record describing the relationship of that prompt with a parent:

```json
{
    "version": 1,
    "record_id": "48803289eb650f8b4b716a5bfb0e24ba",
    "ts": 1788356809,
    "agent": "ACME.UCDUW5V44EBDBIK2FL4CTNDBQFNGBEZVJHSZGQVKRHASN4AV4IWPB5NT",
    "thread_id": "9f2c4b1e8a7d33051c6e0b42d78a91f0",
    "parent_id": "ba939f9408c1f28feb3636ee22044fa4",
    "root_id": "ba939f9408c1f28feb3636ee22044fa4",
    "tool_call_id": "call_1",
    "turn_count_hint": 3
}
```

Edge records are signed and published to a configurable NATS subject (`TRACE.edges` by default). By inspecting the HTTP traffic and the edge records it's possible to reconstruct the whole interaction tree.

Note that this is completely opt-in and backwards compatible. Agents and clients with tracing turned off are indistinguishable from other agents.

PR #199, opened 2026-09-17
ci(claude-code): hold the bundle check's install to bun.lock

M64GitHub:
## Summary

The `agents/claude-code` CI job rebuilds `runtime/server.js` from the packed SDKs and fails on any diff. Its install was `npm install --no-package-lock`, so each run pulled the **newest** external releases, not the versions in `agents/claude-code/bun.lock`. Transitive packages published since the last green rollout run on 2026-09-01 change the minified bundle, so correctly built bundles now fail the check. Example: `sdk-release-rollout` at 287b28b (the #198 merge) rebuilds byte-for-byte with the locked versions but not with these:

| Package | `bun.lock` | Newest |
| --- | --- | --- |
| hono | 4.13.5 | 4.13.8 |
| jose | 6.2.10 | 6.2.12 |
| fast-uri | 3.1.6 | 3.1.8 |
| ip-address | 10.5.0 | 10.7.2 |
| proxy-addr | 2.0.7 | 2.0.8 |
| express-rate-limit | 8.6.2 | 8.7.0 |
| qs | 6.15.3 | 6.16.0 |
| range-parser | 1.2.1 | 1.3.0 |

This PR makes the job install from the lockfile. **It contains no dependency updates and no lock changes.** Whether to do a reviewed refresh of those versions is a separate question.

## Why not `bun install --frozen-lockfile`

Bun can't install this lock as it stands:
- **bun 1.3.14**, the version the job pins and that built the committed bundle, hangs before resolving anything, whether given the packed or the source SDKs. That's the `file:` SDK cycle recorded in the lock.
- **bun 1.4.0** refuses the lock's `file:../../client-sdk/typescript` and `file:../../agent-sdk/typescript` entries ("unsafe folder path"). Moving to it would also change the minified bundle bytes.

Deleting the lock would let Bun install, but that is exactly the unlocked resolution this PR removes.

## Change

`agents/claude-code/scripts/lock-install.ts` wraps the job's existing npm install of the packed SDKs:

1. **`pin`**, before the install, writes npm `overrides` into the CI checkout's `package.json`, one per locked position.
   - A nested version is scoped by its full ancestor path. For example, `eslint/ajv/json-schema-traverse@0.4.1` never overrides the runtime `ajv@8.20.0`'s `json-schema-traverse@1.0.0`.
   - Direct dependencies are already exact and are referenced as `$name`, not repeated.
2. **`verify`**, after the install, reads `node_modules/.package-lock.json` and fails unless every installed external sits at a `bun.lock` position with that position's version and sha512 integrity. Only the packed SDK packages themselves are exempt; anything npm nests below them must also match.

The packed-SDK flow and every later step (bundle build and diff, `verify:bundle`, typecheck, tests, package contents) are unchanged. `README-DEV.md` now describes these steps. It previously pointed to a bundle procedure in `agents/claude-code/README.md` that isn't there.

## Verification (local replay of the job, bun 1.3.14)

- **Replay against 287b28b plus this change:** 437 locked positions pinned; 103 installed externals verified; the rebuilt bundle matches the committed one byte-for-byte (`git diff --exit-code` clean); `verify:bundle`, typecheck, tests (11 unit plus the roundtrip and identity scripts) and `npm pack --dry-run` all pass.
- **Negative check:** without `pin`, `verify` fails and lists exactly the 8 drifted packages above.
- **Placement check:** an earlier draft keyed nested pins by immediate parent only. `json-schema-traverse` then hoisted at 0.4.1, and the rebuilt bundle differed even though every version was in the lock. That's why `verify` now checks positions, not just versions.
- `lock-install.ts` passes a strict typecheck (the package tsconfig's options) with the locked TypeScript 5.9.3.
- `release.py freeze check` matches the baseline; `validate-source` validated 25 manifests; the release-tool unit tests pass (33).

## Which reds this explains

I checked the claude-code job history on the PRs this was expected to explain:

- **#198 (4cdef45):** red for this reason. CI's rebuild differed from the committed bundle, which reproduces byte-for-byte with the `bun.lock` versions.
- **#194:**
  - **2026-09-02 (ba5c417):** red because the bundle hadn't been rebuilt after the SDK changes; fixed by 07f9b33 "Rebuild the Claude Code runtime bundle…".
  - **2026-09-03:** green on all three runs.
  - **Current head 96c1fd1 (red since 2026-09-15):** that commit changes `agent-sdk/typescript/src/service.ts` and client-sdk sources without rebuilding `runtime/server.js`, so this red is a stale bundle, not only external drift. It still needs a rebuild after this PR lands.
- **#196 (d5695e8):** its only run (2026-09-16) **passed** the check. Its committed bundle matches an *unpinned* rebuild, meaning it embeds the newer external versions above rather than the `bun.lock` ones. Under this PR it will go red until it's rebuilt with the locked versions; `mario/tracing-integration` (3c47536) already has such a rebuild. #196's current red checks are `Validate graph and freeze` (from its `agents/openclaw/package.json` addition) and the aggregate check.


M64GitHub:

cozis (approved):

PR #208, opened 2026-09-22
Python host: max_concurrent_prompts serves several prompts at once, off by default

M64GitHub:
Adds `max_concurrent_prompts` to the Python `AgentService`. It lets one instance serve several prompts at once, as the TypeScript host already does. The default is 1, and at 1 nothing changes: `start()` registers the same handler as before, and the SDK awaits each prompt in turn.

## Why

nats-py's micro endpoint awaits its handler for each message in turn, so a Python instance serves one prompt at a time. That creates a deadlock. Suppose agent A, while serving a prompt, prompts agent B, and B prompts A back. A cannot take B's prompt until its own handler returns, so the loop A→B→A waits until the caller's timeout. With the option above 1, it completes.

## Behaviour

- **At 1 (the default):** unchanged, on the same code path. The default-mode tests also pass against the unchanged `service.py`.
- **Above 1:**
  - the endpoint waits for a free slot, then starts the prompt in its own `asyncio` task, in its own copy of the context;
  - with every slot busy, the next prompt waits in the subscription, as it does at 1;
  - a handler's shared state must be safe to interleave at every `await`.
- **`stop()`:** cancels the prompts in flight in both modes. Above 1, it also waits for their tasks, so their terminators are sent before it returns.
- **`$SRV.STATS`:** reports the same things as at 1: the request count, each prompt's time from getting its slot to its terminator, and the errors. nats-py only times its own call to the endpoint handler and offers no public way to adjust stats. So each task updates the endpoint's counters itself when it ends. If a nats-py version lacks those fields, `start()` logs a warning and the stats time only the dispatch. The README says so, and a test pins the behaviour against the locked nats-py.
- Nothing changes on the wire, and no TypeScript source changes.

## Tests

| Suite | Before | After |
| --- | --- | --- |
| client-sdk Python | 520 | 520 |
| agent-service Python | 150 | 172 |

- ruff, `ruff format --check` and mypy are clean.
- The release run's Python steps pass, including the artifact scan.
- The new tests cover the serial default, overlap above 1, waiting at the limit, errors, `stop()` with prompts in flight, validation, per-prompt context, and the A→B→A loop in both modes.

## Follow-up, not in this PR

`agents/deerflow` builds its own `AgentService` and cannot pass the option through yet.


M64GitHub:
Thanks, both checked; no code change.

1. **The count during `stop()`.** Confirmed: nats-py increments `_num_requests` before it awaits the handler. So a prompt cancelled inside `slots.acquire()` stays counted with no time. The dispatcher is awaited serially, though, so at most one prompt can be waiting there; every other one stays in the subscription and is never counted. It is also visible only during shutdown. `Service.stop()` stops the endpoints and then the `$SRV` verb subscriptions, so after `stop()` there are no stats to read. At the default, the prompt cancelled in flight is counted without time too. The difference is one request, briefly, while the service shuts down.
2. **The private fields.** `test_the_prompt_endpoint_stats_time_each_prompt_as_at_the_default` asserts the per-prompt time through `$SRV.STATS`. If a nats-py release renames those fields, the fallback takes over and that test fails. That is the CI signal, and it arrives before the degraded stats could ship.

